PR #131, opened 2024-04-25
feat: RA-78 Refactor PostgresLedgerDataProviderService and extract to LedgerBlockService 

shleger:
…ock* methods to LedgerBlockService


api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
```diff
@@ -536,9 +540,8 @@ public String submitTransaction(String signedTransaction) throws ApiException {
    */
   @Override
   public DepositParameters getDepositParameters() {
-    ProtocolParams protocolParametersFromIndexerAndConfig = ledgerDataProviderService.findProtocolParametersFromIndexerAndConfig();
-    return new DepositParameters(protocolParametersFromIndexerAndConfig.getKeyDeposit().toString(),
-        protocolParametersFromIndexerAndConfig.getPoolDeposit().toString());
+    ProtocolParams pp = protocolParamService.findProtocolParametersFromIndexerAndConfig();
```
BerezinD, line 543: it's better to use the word `params` than `pp`

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/PostgresLedgerDataProviderService.java
```diff
@@ -22,137 +17,21 @@
...
-  }
-
-  private void populateTransactions(List<BlockTx> transactions) {
-    for (BlockTx transaction : transactions) {
-      populateUtxos(transaction.getInputs());
-      populateUtxos(transaction.getOutputs());
-      transaction.setStakeRegistrations(
-          stakeRegistrationRepository.findByTxHash(transaction.getHash())
-              .stream().map(StakeRegistration::fromEntity)
-              .toList()); // TODO Refacotring - do this via JPA
```
BerezinD, line 129: Do we still need this TODO?

shleger: This is deleted code...

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
shleger: It is local var in the method with three lines. In my case there will be no line feed. Let's leave it as it is...

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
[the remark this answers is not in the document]
shleger: I think it is much cleaner:
<pre>

    HttpHeaders headers = new HttpHeaders();
    headers.add(Constants.CONTENT_TYPE_HEADER_KEY, Constants.CBOR_CONTENT_TYPE);
      ResponseEntity<String> exchange = restTemplate.postForEntity(
          submitURL,
          new HttpEntity<>(HexUtil.decodeHexString(signedTransaction), headers),
          String.class);
      if (exchange.getStatusCode().value() == Constants.SUCCESS_SUBMIT_TX_HTTP_CODE) {
        return Optional
            .ofNullable(exchange.getBody())
            .filter(txHash -> txHash.length() == Constants.TX_HASH_LENGTH + 2)
            // removing leading and trailing quotes returned from node API
            .map(txHash -> txHash.substring(1, txHash.length() - 1))
            .orElseThrow(() -> ExceptionFactory.sendTransactionError("Transaction hash is empty"));
      }else {
        log.error("[submitTransaction] There was an error submitting transaction");
        throw ExceptionFactory.sendTransactionError(String.valueOf(exchange.getBody()));
      }

</pre>

BerezinD (approved):

PR #138, opened 2024-05-02
feat: RA-90 Switch to RestTemplate

shleger:


api/src/test/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImplTest.java
```diff
@@ -3,20 +3,32 @@
...
+import java.util.stream.IntStream;
 
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.util.ReflectionTestUtils;
+import org.springframework.web.client.RestTemplate;
 import co.nstant.in.cbor.CborException;
+import com.bloxbean.cardano.client.util.HexUtil;
+import org.jetbrains.annotations.NotNull;
```
BerezinD, line 16: Maybe, jakarta.validation.constraints.NotNull ?

api/src/main/java/org/cardanofoundation/rosetta/api/block/model/domain/Delegation.java
```diff
@@ -6,7 +6,7 @@
 import lombok.NoArgsConstructor;
 
 @Data
-@Builder //TODO saa: refactor tests and remove builder and *argConstructor annotations
```
BerezinD, line 9: Do we remove these TODOs because it's not applicable anymore and the code is fixed?

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/ConstructionApiService.java
```diff
@@ -1,7 +1,21 @@
 package org.cardanofoundation.rosetta.api.construction.service;
 
 
-import org.openapitools.client.model.*;
```
BerezinD, line 4: Good catch!
It's strange that the spotless plugin didn't alert it.

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
```diff
@@ -84,6 +80,7 @@ public class CardanoServiceImpl implements CardanoService {
 
   private final ProtocolParamService protocolParamService;
 
+  private final RestTemplate restTemplate;
```
BerezinD, line 83: Don't want to nitpick, but this final field is better looking on the 82nd line since a block of final fields is shaping.

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
```diff
@@ -368,22 +365,23 @@ public UnsignedTransaction createUnsignedTransaction(NetworkIdentifierType netwo
     String transactionBytes = HexUtil.encodeHexString(
         com.bloxbean.cardano.yaci.core.util.CborSerializationUtil.serialize(mapCbor));
     log.info("[createUnsignedTransaction] Hashing transaction body");
-    String bodyHash = HexUtil.encodeHexString(
+    String bodyHash = com.bloxbean.cardano.client.util.HexUtil.encodeHexString(
```
BerezinD, line 368: Why you are adding here the name of the package? I believe that we have imported exactly the same package (as on the 372nd line)

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
```diff
@@ -408,10 +406,20 @@ public List<SigningPayload> constructPayloadsForTransactionBody(String transacti
...
+        depositParams.getKeyDeposit());
+    Map<String, Double> depositsSumMap = getStringDoubleMap(depositParams, result, refundsSum);
+    long fee = calculateFee(result.getInputAmounts(), result.getOutputAmounts(),
+        result.getWithdrawalAmounts(), depositsSumMap);
+    log.info("[processOperations] Calculated fee:{}", fee);
+    return fillProcessOperationsReturnObject(result, fee);
+  }
+
+  @NotNull
+  private static Map<String, Double> getStringDoubleMap(DepositParameters depositParameters,
```
BerezinD, line 421: Is it, like, `getDepositsSumMap`?

api/src/main/java/org/cardanofoundation/rosetta/api/block/model/domain/PoolRetirement.java
```diff
@@ -6,7 +6,7 @@
 import lombok.NoArgsConstructor;
 
 @Data
-@Builder // TODO saa: remove this and refactor tests
```
Kammerlo, line 9: Same question as Dimitri already had at the previous change.

api/src/main/java/org/cardanofoundation/rosetta/api/block/model/domain/Delegation.java
shleger: Yes, I decided to remove it, because our general approach to use Builders, I need only def and all args constructors, so it is not contradict the model class. After refactoring @Builders is not required but tests are already written with it...

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
shleger: I will change, maybe auto_magic)

api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
shleger: Yes, will rename, thanks!

api/src/main/java/org/cardanofoundation/rosetta/api/block/model/domain/PoolRetirement.java
shleger: Answered

BerezinD (approved):

Kammerlo (approved):
Could you fix the new sonar issues, then it's fine

PR #141, opened 2024-05-03
Feat/ra 51 verify network in request

Kammerlo:


api/src/main/java/org/cardanofoundation/rosetta/api/network/service/impl/NetworkServiceImpl.java
```diff
@@ -199,4 +199,43 @@ private List<Producer> getPublicRoots(List<PublicRoot> publicRoots) {
...
+      case Constants.PREPROD_PROTOCOL_MAGIC -> {
+        return network.equals(Constants.PREPROD);
+      }
+      case Constants.PREVIEW_PROTOCOL_MAGIC -> {
+        return network.equals(Constants.PREVIEW);
+      }
+      case Constants.DEVKIT_PROTOCOL_MAGIC -> {
+        return network.equals(Constants.DEVKIT);
+      }
+      default -> throw ExceptionFactory.networkNotFoundError();
```
BerezinD, line 237: Considering returning a boolean, it's not clear that here could an exception be thrown. 
Let's either throw exceptions within verify methods or just in the initial one. 
*the first option will be clearer since the method `verifyNetworkRequest` returns void.

BerezinD (approved):

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/impl/NetworkServiceImpl.java
```diff
@@ -199,4 +199,45 @@ private List<Producer> getPublicRoots(List<PublicRoot> publicRoots) {
...
+      }
+      case Constants.DEVKIT_PROTOCOL_MAGIC -> {
+        return network.equals(Constants.DEVKIT);
+      }
+      default -> {
+        return false;
+      }
+    }
+  }
+
```
shleger, line 242: I cannot find unit tests for this code

Kammerlo: Added Tests

PR #142, opened 2024-05-03
refactor: fix sonarcloud code smells for injection issues

Kammerlo:
I started fixing some sonarcloud issues to get the number down in our dashboard.


api/src/main/java/org/cardanofoundation/rosetta/api/block/mapper/StakeRegistrationToOperation.java
```diff
@@ -40,10 +40,12 @@ private String convert(CertificateType model) {
     if (model == null) {
       return null;
     } else {
-      return model.equals(CertificateType.STAKE_REGISTRATION)
-          ? OperationType.STAKE_KEY_REGISTRATION.getValue() :
-          model.equals(CertificateType.STAKE_DEREGISTRATION)
-              ? OperationType.STAKE_KEY_DEREGISTRATION.getValue() : null;
+      if(model.equals(CertificateType.STAKE_REGISTRATION)) {
```
BerezinD, line 43: Great! And it could be even better with the switch:
`return switch (model) {
      case STAKE_REGISTRATION -> OperationType.STAKE_KEY_REGISTRATION.getValue();
      case STAKE_DEREGISTRATION -> OperationType.STAKE_KEY_DEREGISTRATION.getValue();
      default -> null;
    };`

api/src/main/java/org/cardanofoundation/rosetta/common/interceptor/CustomResponseBodyAdviceAdapter.java
```diff
@@ -14,12 +15,11 @@
...
 
 @ControllerAdvice
+@RequiredArgsConstructor
 public class CustomResponseBodyAdviceAdapter implements ResponseBodyAdvice<Object> {
 
-    @Autowired
-    LoggingService loggingService;
+    final LoggingService loggingService;
 
-    @Override
```
BerezinD, line 22: Are you sure that it's not overriding (implementing) a parent method anymore? 🤔

BerezinD (approved):

api/src/main/java/org/cardanofoundation/rosetta/common/interceptor/CustomResponseBodyAdviceAdapter.java
Kammerlo: Good point

PR #151, opened 2024-05-07
feat:code-coverage

AlksndrSlgb:


pom.xml
```diff
@@ -54,6 +54,10 @@
         <version.modelmapper>3.2.0</version.modelmapper>
         <version.assertj-core>3.25.3</version.assertj-core>
         <version.h2>2.2.224</version.h2>
+        <sonar.coverage.jacoco.xmlReportPaths>
+        ./target/site/
+        jacoco-aggregate/jacoco.xml
```
BerezinD, line 59: Are you sure that we have the `jacoco-aggregate` folder and jacoco.xml in it? Also, one line would be better :)
`target/site/jacoco/jacoco.xml`

AlksndrSlgb: It's just draft :)))

PR #152, opened 2024-05-08
RA-92: Changed output data from Double to Integer to maintain consistency

Gennadii-T:


api/src/main/java/org/cardanofoundation/rosetta/common/services/impl/CardanoServiceImpl.java
```diff
@@ -220,7 +220,7 @@ public Double calculateTxSize(NetworkIdentifierType networkIdentifierType,
 
     String transaction = buildTransaction(unsignedTransaction.bytes(), signaturesList,
         unsignedTransaction.metadata());
-    return ((double) transaction.length() / 2);
+    return (transaction.length() / 2);
```
shleger, line 223: Why we divide transaction size and what happens if tx size is odd?

Gennadii-T: @Kammerlo could you please clarify this? I'm not very familiar with the original logic

shleger: As in the return statement `(transaction.length() / 2);` we should divide transaction length on two, why?
And why we can replace double with int in this case?

Kammerlo: It's not quite obvious. But the String represents Hexdata. So each Charakter is one hex char which is 4Bit. To get the size in bytes you have to divide the hexstring by two. I'll add a comment in your Branch to make it clearer for the next one to stumble across this line

Kammerlo (approved):

PR #160, opened 2024-05-09
RA-87 fix docker compose build

MarMarMot:


Kammerlo (changes requested):
As in your other PR. Please change the folder where the configs are stored from current to config. 
It's not common to name it "current", because there are configs in there.

Kammerlo (approved):

PR #165, opened 2024-05-15
fix: RA-126 model mapper storage

shleger:


Kammerlo (approved):
@Gennadii-T Could you test this if it fixes the bug? If yes it's fine and can be merged

Gennadii-T (changes requested):
I need to discuss the result because I'm still seeing the same issue

shleger:
Will switch to another mapper lib

PR #167, opened 2024-05-15
chore: added caching of genesis block, since it doesn't change at all.

Kammerlo:
As part of optimization. The Genesis block doesn't change, so we should cache it.


Kammerlo:
This halved the response time for e.g. /network/status from 1.3 seconds to 0.6s

api/src/main/java/org/cardanofoundation/rosetta/api/block/service/LedgerBlockServiceImpl.java
```diff
@@ -113,10 +115,13 @@ public Block findLatestBlock() {
 
   @Override
   public GenesisBlock findGenesisBlock() {
-    log.debug("About to run findGenesisBlock query");
-    return blockRepository.findGenesisBlock()
-        .map(b -> mapper.map(b, GenesisBlock.class))
-        .orElseThrow(ExceptionFactory::genesisBlockNotFound);
+    if(cachedGenesisBlock == null) {
```
shleger, line 118: Maybe introduce caching here?
https://stackoverflow.com/a/26283080/25051587
In this case in theory we can reset caches with actuator..

shleger (approved):

PR #169, opened 2024-05-15
Changing packages structure, cleaning up JPA entities

BerezinD:


api/src/main/java/org/cardanofoundation/rosetta/common/services/ProtocolParamServiceImpl.java
```diff
@@ -0,0 +1,40 @@
...
+@Slf4j
+@RequiredArgsConstructor
+public class ProtocolParamServiceImpl implements ProtocolParamService {
+
+  private final EpochParamRepository epochParamRepository;
+  private final ProtocolParamsToEntity mapperProtocolParams;
+
+  private ProtocolParams cachedProtocolParams;
+
+  @PostConstruct
```
Kammerlo, line 25: This PostConstruct leads to a crash of the intregrationtest. I had the same problem with the genesisblock caching. 
A work-around would be to save not on postconstruct, but when it is queried the first time.

Kammerlo (approved):

api/src/main/java/org/cardanofoundation/rosetta/common/services/ProtocolParamServiceImpl.java
BerezinD: Thank you for the information!
I've added some old-style thread-safe code to get parameters on the first call:
https://github.com/cardano-foundation/cardano-rosetta-java/pull/169/files#diff-e1ab139173b73f29bec8b97daa5b2b6d8a04b47a287be793fa90672656284344R25

api/src/main/java/org/cardanofoundation/rosetta/common/mapper/DataMapper.java
```diff
@@ -52,20 +50,23 @@
 public class DataMapper {
```
Gennadii-T, line 50: Could you put it in separate methods Currency.builder() and Amount.builder() to maintain readability and avoid code duplication

Gennadii-T (approved):

AlksndrSlgb (approved):

api/src/main/java/org/cardanofoundation/rosetta/common/mapper/DataMapper.java
BerezinD: Do you mean moving the default currency to a separate method?
`Currency.builder()
        .decimals(Constants.ADA_DECIMALS)
        .symbol(Constants.ADA).build();`

shleger (approved):

api/src/main/java/org/cardanofoundation/rosetta/common/mapper/DataMapper.java
BerezinD: I will refactor it with the next PR, thanks for the comment.

PR #182, opened 2024-05-19
Feat/ra 133 optimize mappers

Kammerlo:


api/src/main/java/org/cardanofoundation/rosetta/api/block/mapper/BlockTxToRosettaTransaction.java
```diff
@@ -1,79 +1,102 @@
...
-  final PoolRegistrationToOperation poolRegistrationToOperation;
-  final WithdrawalToOperation withdrawalToOperation;
-
-  private static final OperationStatus status = OperationStatus.builder()
+@Mapper(config = BaseMapper.class, uses = {InputToOperation.class, StakeRegistrationToOperation.class,
+    DelegationToOperation.class, PoolRegistrationToOperation.class, PoolRetirementToOperation.class,
+    WithdrawalToOperation.class, OutputToOperation.class})
+public abstract class BlockTxToRosettaTransaction {
+
+  @Autowired
```
Kammerlo, line 28: Mapstruct currently doesn't support field injection when generating code

pom.xml
```diff
@@ -54,6 +54,8 @@
         <version.modelmapper>3.2.0</version.modelmapper>
```
shleger, line 54: Why modelmapper dependency does not removed?

shleger (approved):

AlksndrSlgb (approved):

pom.xml
Kammerlo: I'll remove it today or latest tomorrow morning, since there are also a lot of tests to be adjusted.

PR #183, opened 2024-05-20
RA-120 improve network status endpoint performance

MarMarMot:


api/src/main/java/org/cardanofoundation/rosetta/api/network/service/NetworkServiceImpl.java
```diff
@@ -236,10 +195,36 @@ private boolean verifyNetwork(String network) {
...
     }
   }
 
+  public Integer getNetworkMagic() {
+    Integer magicNumber = cachedMagicNumber;
+    if (magicNumber == null) {
+      synchronized (lock) {
+        magicNumber = cachedMagicNumber;
+        if (magicNumber == null) {
+          JSONObject json = loadGenesisShelleyConfig();
```
BerezinD, line 213: Can we do it in the @PostConstruct?

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/TopologyConfigServiceImpl.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public List<Peer> getPeers() {
+    List<Peer> peers = cachedPeers;
+    if (peers == null) {
+      synchronized (lock) {
+        peers = cachedPeers;
+        if (peers == null) {
+          TopologyConfig topologyConfig = loadTopologyConfig();
```
BerezinD, line 46: If we get peers only from the file, it's better to make it initialized with @PostConstruct. Or don't we have the file yet after bean initialization?

api/src/main/java/org/cardanofoundation/rosetta/common/util/RosettaConstants.java
```diff
@@ -150,6 +150,8 @@ public enum RosettaErrorType {
     OUTSIDE_VALIDITY_INTERVAL_UTXO(
         "Error when sending the transaction - OutsideValidityIntervalUTxO", 4037),
     CONFIG_NOT_FOUND("Environment configurations needed to run server were not found", 4038),
+    LATEST_BLOCK_NOT_FOUND("Сan't get the latest block. The node has not yet been initialized",
```
BerezinD, line 153: If we can't find any block (because the node has not started yet), then it's an error `GENESIS_BLOCK_NOT_FOUND`. Else, it's just some block that could not be found.
I believe that we could omit a new error creation.

api/src/main/java/org/cardanofoundation/rosetta/api/block/model/entity/BlockEntity.java
```diff
@@ -58,4 +58,10 @@ public class BlockEntity {
 
   @Column(name = "epoch")
   private Integer epochNumber;
+
+  public BlockEntity(String hash, Long number, Long blockTimeInSeconds) {
```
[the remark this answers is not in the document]
MarMarMot: I think projection is the better solution
Done

api/src/main/java/org/cardanofoundation/rosetta/api/block/model/entity/projection/BlockIdentifierProjection.java
```diff
@@ -0,0 +1,8 @@
+package org.cardanofoundation.rosetta.api.block.model.entity.projection;
+
+public interface BlockIdentifierProjection {
```
BerezinD, line 3: Just FYI, it could be significantly slow:
https://stackoverflow.com/questions/53347063/why-are-interface-projections-much-slower-than-constructor-projections-and-entit

Let's just keep that in mind during the performance results review

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/NetworkServiceImpl.java
```diff
@@ -57,32 +50,25 @@
...
-  private final ResourceLoader resourceLoader;
 
   @PostConstruct
-  public void filePathExistingValidator() {
-    validator(topologyFilepath);
-    validator(genesisPath);
+  public void init() {
+    JSONObject json = loadGenesisShelleyConfig();
+    cachedMagicNumber = (Integer) json.get(Constants.NETWORK_MAGIC_NAME);
+    log.info("Magic number loaded from genesis config json");
```
BerezinD, line 69: Let's add the number itself

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/NetworkServiceImpl.java
```diff
@@ -163,47 +146,24 @@ public Network getSupportedNetwork() {
...
-    Block latestBlock = ledgerBlockService.findLatestBlock();
-    log.debug("[networkStatus] Latest block found {}", latestBlock);
+    BlockIdentifierExtended latestBlock = ledgerBlockService.findLatestBlockIdentifier();
+    log.debug("[networkStatus] Latest block found " + latestBlock);
+
     log.debug("[networkStatus] Looking for genesis block");
-    GenesisBlock genesisBlock = ledgerBlockService.findGenesisBlock();
-    log.debug("[networkStatus] Genesis block found {}", genesisBlock);
+    BlockIdentifierExtended genesisBlock = ledgerBlockService.findGenesisBlockIdentifier();
+    log.debug("[networkStatus] Genesis block found " + genesisBlock);
```
BerezinD, line 156: Did you change the parametrized string intentionally?

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/TopologyConfigServiceImpl.java
MarMarMot: Done

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/NetworkServiceImpl.java
MarMarMot: Done

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/NetworkServiceImpl.java
MarMarMot: No, fixed

api/src/main/java/org/cardanofoundation/rosetta/api/network/service/NetworkServiceImpl.java
MarMarMot: Done

Kammerlo (approved):

PR #185, opened 2024-05-20
RA-105: Increasing code coverage

AlksndrSlgb:


pom.xml
```diff
@@ -54,6 +54,24 @@
...
+            **/org/openapitools/**,
+            **/model/**,
+            **/entity/**,
+            **/rosetta/config/**,
+            **/rosetta/common/enumeration/NetworkId.java,
+            **/rosetta/common/enumeration/OperationTypeStatus.java,
+            **/rosetta/common/enumeration/RewardType.java,
+            **/rosetta/common/enumeration/SyncStateType.java,
+            **/rosetta/common/enumeration/ScriptPurposetype.java,
+            **/rosetta/api/mempool/service/MempoolServiceImpl.java,
```
Kammerlo, line 69: Could you add a comment here, that says something like "excluded for now, will be integrated when Mempool is implemented completely" just as an example :)

Kammerlo (approved):

PR #189, opened 2024-05-22
Chore: Artillery target port fixed

MarMarMot:


Kammerlo (approved):

artilleryTests/rosetta-load-test.yml
```diff
@@ -1,5 +1,7 @@
 config:
-  target: http://localhost:8081
+  target: http://127.0.0.1:8082
+  ensure:
+    maxErrorRate: 1
   http:
     defaults:
```
BerezinD, line 6: `timeout: 60` should be under the `http:`, not `http: http:`

artilleryTests/rosetta-load-test.yml
```diff
@@ -21,12 +23,13 @@ config:
       name: Spike phase
     - duration: 60
```
BerezinD, line 24: This one is absolete

BerezinD:
Good job!
But we still have 
`plugins.metrics-by-endpoint./block.codes.500: .................................. 1230`
Probably, because of the index.

artilleryTests/rosetta-load-test.yml
MarMarMot: Fixed

artilleryTests/rosetta-load-test.yml
MarMarMot: Fixed

Kammerlo (approved):

PR #194, opened 2024-05-24
Account performance optimizations

BerezinD:


Kammerlo (approved):

Gennadii-T (approved):

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/AccountServiceImpl.java
```diff
@@ -67,9 +81,9 @@ public AccountCoinsResponse getAccountCoins(AccountCoinsRequest accountCoinsRequ
     }
     log.debug("[accountCoins] Address is Era");
     if (Objects.nonNull(currencies)) {
```
shleger, line 83: Can be rewritten wit Optional.ofNullable(..)

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/AccountServiceImpl.java
```diff
@@ -84,33 +98,60 @@ private AccountBalanceResponse findBalanceDataByAddressAndBlock(String address,
...
+  private boolean isTokenNameValid(String name) {
+    return TOKEN_NAME_VALIDATION.matcher(name).matches() || isEmptyHexString(name);
+  }
+
+  private boolean isPolicyIdValid(String policyId) {
+    return POLICY_ID_VALIDATION.matcher(policyId).matches();
+  }
+
+  private List<Currency> filterRequestedCurrencies(List<Currency> currencies) {
+    if (currencies != null && currencies.stream().map(Currency::getSymbol)
```
shleger, line 150: Also, can be refactored in functional style (without if(..))

api/src/main/java/org/cardanofoundation/rosetta/common/util/CardanoAddressUtils.java
```diff
@@ -169,11 +169,13 @@ public static EraAddressType getEraAddressType(String address) {
...
       if (address.startsWith(Constants.ADDRESS_PREFIX)
           || address.startsWith(StakeAddressPrefix.MAIN.getPrefix())) {
+        // validate bech32 address. Unfortunately, it will throw a runtime exception in case of invalid address
+        Bech32.decode(address);
         return EraAddressType.SHELLEY;
       }
       new ByronAddress(address).getAddress();
       return EraAddressType.BYRON;
-    } catch (AddressFormatException | AddressRuntimeException e) {
+    } catch (RuntimeException e) {
```
shleger, line 178: Need to replace with custom ApiException

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/LedgerAccountServiceImpl.java
```diff
@@ -15,22 +18,21 @@
...
-import org.cardanofoundation.rosetta.api.block.model.domain.StakeAddressBalance;
-import org.cardanofoundation.rosetta.api.block.model.repository.StakeAddressRepository;
+import org.cardanofoundation.rosetta.api.account.model.repository.StakeAddressBalanceRepository;
 import org.cardanofoundation.rosetta.common.util.Formatters;
 
 @Slf4j
 @RequiredArgsConstructor
 @Component
-@Transactional(readOnly = true)
+@Transactional(readOnly = true, propagation = Propagation.NEVER)
```
shleger, line 30: We cannot use this method inside a transaction. Maybe better just leave   propagation by default.

shleger (approved):

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/LedgerAccountServiceImpl.java
BerezinD: This will affect only the Account service method. Let's keep it for these changes, I will thoroughly test it with the next PR and remove it or leave it with a comment.

api/src/main/java/org/cardanofoundation/rosetta/common/util/CardanoAddressUtils.java
BerezinD: Unfortunately, `Bech32.decode(address)` will throw a runtime exception in case of an invalid address. 
`public static Bech32Data decode(String bech32EncodedString) {
    Tuple<String, byte[]> bech32Data = bech32Decode(bech32EncodedString);
    String hrp = (String)bech32Data._1;
    byte[] b32Arr = (byte[])bech32Data._2;
    if (b32Arr.length < 6) {
      throw new RuntimeException("Invalid data length.");
    } else if (!verifyChecksum(hrp, b32Arr)) {
      throw new RuntimeException("Invalid checksum.");
    } else {
      byte[] data = Arrays.copyOfRange(b32Arr, 0, b32Arr.length - 6);
      byte[] b256Arr = convertBits(data, 5, 8, false);
      if (b256Arr == null) {
        throw new RuntimeException("Invalid data format.");
      } else {
        byte witVer = b32Arr[0];
        return new Bech32Data(hrp, b256Arr, witVer);
      }
    }
  }`
  
  We cannot change it here.

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/AccountServiceImpl.java
BerezinD: That's reasonable. Done.

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/AccountServiceImpl.java
BerezinD: This will add complexity with almost no improvements. Let's keep it with a simple if() statement.
https://stackoverflow.com/questions/52038417/should-optional-ofnullable-be-used-for-null-check

PR #200, opened 2024-05-28
Adding info endpoint

AlksndrSlgb:


Kammerlo (approved):
Ah this one returns the git hash?

AlksndrSlgb:
> Ah this one returns the git hash?
Not only git hash, but also some information about build
In my branch I get 
{
    "git": {
        "branch": "feat/Adding-info-endpoint",
        "commit": {
            "id": "3d40304",
            "time": "2024-05-28T18:13:15Z"
        }
    },
    "build": {
        "artifact": "api",
        "name": "api",
        "time": "2024-05-29T07:25:05.864Z",
        "version": "1.0.0-SNAPSHOT",
        "group": "org.cardanofoundation.rosetta-java"
    }
}

api/src/main/resources/config/application.yaml
```diff
@@ -48,3 +48,9 @@ logging:
     name: ${LOG_FILE_NAME:rosetta-api.log}
     max-size: ${LOG_FILE_MAX_SIZE:10MB}
     max-history: ${LOG_FILE_MAX_HISTORY:10}
+
+management:
+  endpoints:
+    web:
+      exposure:
+        include: "info"
```
BerezinD, line 56: Will this enable **only** "info" endpoints? What are those? We need to add them to the documentation.

AlksndrSlgb: Only one info endpoint with information which I posted before

BerezinD: So, will we delete it for the production? If so, you could place it into the dev environment profile.

PR #206, opened 2024-05-31
RA-145: Add cache for ProtocolParamService

Gennadii-T:


api/src/main/java/org/cardanofoundation/rosetta/config/CacheConfig.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.cache.concurrent.ConcurrentMapCacheManager;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+@Configuration
+@EnableCaching
+@EnableScheduling
```
BerezinD, line 11: Do we need here the `@EnableScheduling` annotation?

Gennadii-T: Actually, you're right. We already have the @EnableScheduling annotation on the RosettaApiApplication class.

BerezinD (approved):

PR #207, opened 2024-05-31
RA-145: Added trigger-based processing for updates to the 'epoch_param' table

Gennadii-T:


Kammerlo (commented):
Code is fine, we still need to discuss which approach is the best for us.

PR #211, opened 2024-06-02
RA-139: Optimized construction combine

Gennadii-T:


Kammerlo (approved):

Kammerlo:
If all tests run, then it's fine from my side. Thanks!

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java
```diff
@@ -288,11 +276,13 @@ public TransactionWitnessSet getWitnessesForTransaction(List<Signatures> signatu
     ArrayList<BootstrapWitness> bootstrapWitnesses = new ArrayList<>();
     log.info("[getWitnessesForTransaction] Extracting witnesses from signatures");
     signaturesList.forEach(signature -> {
-      VerificationKey vKey = new VerificationKey();
-      vKey.setCborHex(ObjectUtils.isEmpty(signature) ? null : signature.publicKey());
+      if (ObjectUtils.isEmpty(signature)) {
```
BerezinD, line 279: `Signature` is a record and not an array of any other child of Iterable. Thus, it cannot be empty - only null or not. Probably, we should check the `signature` field itself (or else an NPE will be thrown later).

BerezinD (approved):

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java
BerezinD: I will check validations in the scope of the next sub-task.
https://cardanofoundation.atlassian.net/browse/RA-139

PR #212, opened 2024-06-03
RA-147: Improve Artillery Tests

AlksndrSlgb:


Kammerlo (approved):

artilleryTests/rosetta-load-test.yml
```diff
@@ -1,5 +1,16 @@
 config:
   target: http://127.0.0.1:8082
+  payload:
+    path: 'data.csv'
+    fields:
+      - 'address'
+      - 'block'
```
BerezinD, line 7: Now we have the 0th, 50th, 29th, and 22nd blocks to be checked. Do they have any transactions?

BerezinD (approved):

BerezinD:
We need to increase the number of virtual users
image

artilleryTests/rosetta-load-test.yml
AlksndrSlgb: Good question! The 50th block has, the others do not.
I'll change it.

PR #221, opened 2024-06-12
Refactoring for code coverage

AlksndrSlgb:


yaci-indexer/pom.xml
```diff
@@ -19,11 +19,6 @@
         <yaci-store.version>0.1.0-rc2-323a1a4-SNAPSHOT</yaci-store.version>
         <sonar.exclusions>src/main/java/org/cardanofoundation/rosetta/yaciindexer/stores/txsize/model/*</sonar.exclusions>
         <version.spotless-maven-plugin>2.43.0</version.spotless-maven-plugin>
-        <test.exclusions>
-            **/model/**,
-            **/yaciindexer/config/**,
```
Kammerlo, line 24: Don't we need to exclude this? Because we don't have tests for that and don't want to write tests for this?

AlksndrSlgb: It's already added to base pom.xml 
` **/yaciindexer/**`

AlksndrSlgb: now we have a lot of uncovered classes in sonar because of this reason
image

Kammerlo: Ah ok!

Kammerlo (approved):

PR #223, opened 2024-06-16
Chore: create index for table address_utxo for column tx_hash if not …

MarMarMot:
…exists wher docker file running


Kammerlo (approved):

docker/.env.dockerfile
```diff
@@ -6,7 +6,7 @@ PROTOCOL_MAGIC=1
 # mainnet 764824073, preprod 1, preview 2, sanchonet 4, devkit 42
 
 # Node synchronization
-SYNC=true
+SYNC=false
```
BerezinD, line 9: Why are we disabling node synchronization?

MarMarMot: fixed

PR #224, opened 2024-06-18
Chore: Default node/db dirs creation added in dockerfile

MarMarMot:


docker/Dockerfile
```diff
@@ -181,6 +181,7 @@ COPY ./config /networks
 RUN mkdir /config
 
 RUN mkdir -p /node/db
+RUN mkdir -p /node/postgres
```
BerezinD, line 184: What are we adding there? Is this folder necessary?

Kammerlo: yes it's needed otherwise we have permission issues

Kammerlo (approved):

BerezinD (approved):

PR #227, opened 2024-06-28
correct genesis + switch Dockerfile to mainnet

Godspeed-exe:


Kammerlo:
Merging into this repository so that I'm able to continue from here. 
Thanks for update this really helps @Godspeed-exe !

PR #233, opened 2024-07-09
chore: upgrade to node 9.1

Kammerlo:
This PR is for keeping up everyone involved updated.

- Initial work done for upgrading to 9.0
- Testings currently running, I'll ping you, when I'm done.


Godspeed-exe:
Keep into account that we still need the new Conway file to be fully ready for the Chang HF. @Kammerlo

PR #273, opened 2025-01-31
feat: withdrawable rewards should be available on Rosetta's account/b…

### LINKED ISSUE #272: Add Rewards data to rosetta-java
Currently in rosetta-java it is not possible to query pending rewards from a stake address. 
The balance is currently calculated by the utxo. Since it is possible now to query the rewards from a stake address, it must be included in rosetta. 

Currently we are returning the UTxOs owned by the stakeaddress, which is basically wrong in this case. A user expects the pending rewards to be returned, when querying the stakeaddress.

Places to change it:
- [AccountServiceImpl.java](https://github.com/cardano-foundation/cardano-rosetta-java/blob/28bcfa66bdb5e7b4f8e444d2bdcf17d4a10516b0/api/src/main/java/org/cardanofoundation/rosetta/api/account/service/AccountServiceImpl.java#L105)
- Yaci-indexer - Add Rewards quering (an example how to do it [Yaci-Store AccountService](https://github.com/bloxbean/yaci-store/blob/8d3102caed983095feb124d81ebea8c10e5bfb91/aggregates/account/src/main/java/com/bloxbean/cardano/yaci/store/api/account/service/AccountService.java#L26)
The n2c communication between yaci-store and the node is already setup. 

Think of a way how to get the data from yaci-store to the API. 
Probably the best way would be via an http, since we need the response synchronously. 
Possible ways could be establish an http connection or do it via the database.

matiwinnetou:
…alance endpoint for a given stake address.

Fixes #272


api/src/main/java/org/cardanofoundation/rosetta/api/account/mapper/StakeAccountInfoToAddressBalanceMapperImpl.java
```diff
@@ -0,0 +1,28 @@
...
+@Slf4j
+public class StakeAccountInfoToAddressBalanceMapperImpl implements StakeAccountInfoToAddressBalanceMapper {
+
+    @Override
+    public AddressBalance convertToAddressBalance(StakeAccountInfo stakeAccountInfo, Long number) {
+        log.debug("Mapping stake address to address balance: {}", stakeAccountInfo.getStakeAddress());
+
+        return AddressBalance.builder()
+                .address(stakeAccountInfo.getStakeAddress())
+                .unit(ADA)
```
Kammerlo, line 22: If you add ADA here as constant you could set the number as constant 6 as well, then I'd rename. this function to something like `convertToAdaAddressBalance`. Just that it's clear that this will map directly to ADA.
If not I would add the unit to a function parameter as well

api/src/main/java/org/cardanofoundation/rosetta/api/account/service/LedgerAccountServiceImpl.java
```diff
@@ -37,11 +39,16 @@ public List<AddressBalance> findBalanceByAddressAndBlock(String address, Long nu
   }
 
   @Override
-  public List<AddressBalance> findBalanceByStakeAddressAndBlock(String stakeAddress,
-      Long number) {
+  @Deprecated // dead code???, only used in tests
```
Kammerlo, line 42: Yes this isn't used anymore after this merge, can be removed.

api/src/main/java/org/cardanofoundation/rosetta/client/YaciHttpGatewayImpl.java
```diff
@@ -0,0 +1,62 @@
...
+    @Value("${cardano.rosetta.YACI_HTTP_BASE_URL}")
+    private String yaciBaseUrl;
+
+    @Value("${cardano.rosetta.HTTP_REQUEST_TIMEOUT}")
+    private int httpRequestTimeout;
+
+    @Override
+    public StakeAccountInfo getStakeAccountInfo(String stakeAddress) {
+        var getStakeAccountDetailsHttpRequest = HttpRequest.newBuilder()
+                .uri(URI.create(yaciBaseUrl + "/accounts/" + stakeAddress))
```
Kammerlo, line 35: Is this endpoint directly from yaci-store? Or is there anything we need to add in this component as well?

If it's up to us I would prefer renaming this endpoint to rewards or something, but we can discuss offline.

matiwinnetou: yes directly from yaci-store, ok I can rename

PR #276, opened 2025-02-07
feat: stake account balance.

matiwinnetou:


Kammerlo (approved):

api/src/main/resources/config/application.yaml
```diff
@@ -33,6 +33,11 @@ cardano:
     DEVKIT_URL: ${DEVKIT_URL:yaci-cli}
     DEVKIT_PORT: ${DEVKIT_PORT:3333}
     SEARCH_PAGE_SIZE: ${SEARCH_PAGE_SIZE:10}
+    OFFLINE_MODE: ${OFFLINE_MODE:false}
+
+    YACI_HTTP_BASE_URL: ${YACI_HTTP_BASE_URL}
```
Kammerlo, line 38: You either need to define a default value or add it to all .env files otherwise it will lead to problems. 
Additionally this must be overwritten by the single docker mode (this is what exchanges use the most). In this case the env must be set to `localhost:9095`

matiwinnetou: I will, I just ran into issues and removed it.

.env.IntegrationTest
```diff
@@ -96,3 +97,8 @@ LOG_FILE_PATH=/var/log/rosetta-java
 LOG_FILE_NAME=/var/log/rosetta-java/rosetta-java.log
 LOG_FILE_MAX_SIZE=10MB
 LOG_FILE_MAX_HISTORY=10
+
+YACI_HTTP_BASE_URL=http://localhost:9095/api/v1
```
Kammerlo, line 101: in docker compose this url is `http://yaci-indexer:9095` right?

.env.docker-compose
```diff
@@ -60,6 +60,7 @@ INDEXER_DOCKER_IMAGE_TAG=main
 PRUNING_ENABLED=false
 
 YACI_SPRING_PROFILES=postgres,n2c-socket
+YACI_INDEXER_PORT=9095
```
Kammerlo, line 63: If you don't need to use the default value of `YACI_HTTP_BASE_URL` it must be defined here as well

.env.IntegrationTest
```diff
@@ -96,3 +97,8 @@ LOG_FILE_PATH=/var/log/rosetta-java
...
+HTTP_CONNECT_TIMEOUT_SECONDS=5
```
Kammerlo, line 102: I wouldn't name these variables in the env files, since users won't change them and stick to the default values

api/src/main/resources/config/application-offline.yaml
```diff
@@ -11,4 +11,7 @@ spring:
 
 cardano:
   rosetta:
-    OFFLINE_MODE: true
\ No newline at end of file
+    OFFLINE_MODE: true
+    YACI_HTTP_BASE_URL: ${YACI_HTTP_BASE_URL:http://localhost:9095}
+    HTTP_CONNECT_TIMEOUT_SECONDS: ${HTTP_CONNECT_TIMEOUT_SECONDS:5}
```
Kammerlo, line 16: here you can replace ` ${HTTP_CONNECT_TIMEOUT_SECONDS}`with just the value 5. Since users won't be changing this and if they need to they can just use this environment variable `CARDANO_ROSETTA_HTTP_CONNECT_TIMEOUT_SECONDS`

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/service/AccountServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+                return Optional.empty();
+            }
+
+            Address address = new Address(stakeAddress);
+
+            try {
+                localStateQueryClient.release().block(Duration.ofSeconds(5));
+            } catch (Exception e) {
+                //Ignore the error
```
Kammerlo, line 57: Is there a reason why we can ignore this error?

Kammerlo (approved):
From my point of view just minor stuff, Thanks @matiwinnetou !

PR #286, opened 2025-02-18
fix: set transaction TTL based on time with shelly date fallback

matiwinnetou:


matiwinnetou:
The issue you are facing is unrelated to the PR itself.

api/src/main/java/org/cardanofoundation/rosetta/RosettaApiApplication.java
```diff
@@ -29,6 +31,7 @@
     "org.cardanofoundation.rosetta.api.common.model.entity"})
 @OpenAPIDefinition(info = @Info(title = "APIs", version = "1.0", description = "Rosetta APIs v1.0"))
 @EnableScheduling
+@Slf4j
```
Kammerlo, line 34: Not needed here.

api/src/main/java/org/cardanofoundation/rosetta/common/time/OfflineSlotServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.cardanofoundation.conversions.CardanoConverters;
+import org.cardanofoundation.rosetta.common.exception.ApiException;
+import org.cardanofoundation.rosetta.common.exception.ExceptionFactory;
+
+@Service
+@Slf4j
+public class OfflineSlotServiceImpl implements OfflineSlotService {
+
+    @Autowired
```
Kammerlo, line 21: Why do use Autowire here and not constructor injection?

api/src/main/java/org/cardanofoundation/rosetta/common/time/OfflineSlotServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    @Qualifier("shelleyStartSlot")
+    @SuppressWarnings("sonar:RedundantQualifierAnnotation")
+    protected Long shelleyStartSlot;
+
+    @Autowired(required = false)
+    @Nullable
+    protected CardanoConverters cardanoConverters;
+
+    @Override
+    public long getCurrentSlotBasedOnTime() {
```
Kammerlo, line 42: That's a nice solution to estimate the solution. I like it!

Kammerlo (approved):
It looks good! 
Except the sonar issues, but it marked the same like I did. Autowire, unused imports and a public modifier in a test class. So just minor stuff.

PR #291, opened 2025-02-25
chore: docker standard env fixes.

matiwinnetou:


.env.docker-compose
```diff
@@ -75,8 +75,13 @@ LOG_FILE_NAME=/var/log/rosetta-java/rosetta-java.log
 LOG_FILE_MAX_SIZE=10MB
 LOG_FILE_MAX_HISTORY=10
 
+YACI_HTTP_BASE_URL=http://localhost:9095/api/v1
```
Kammerlo, line 78: For the docker compose setup the hostname is not localhost. The communication is done via the docker network, where the container names are used as hostnames. So the hostname here must be `yaci-indexer` otherwise the api component won't be able to connect to the indexer

Kammerlo (changes requested):

Kammerlo (approved):

PR #316, opened 2025-03-07
feat: simple performance / scalability fixes.

### LINKED ISSUE #315: show sql logging is an anti-pattern in high concurrency services
We should remove show sql in log statements, we can introduce a config flag in the future but show sql is anti-pattern for high load scenarios, this is typically a development flag.

matiwinnetou:
Fixes #315


Kammerlo (approved):

api/src/main/resources/config/application-online.yaml
```diff
@@ -4,20 +4,22 @@ spring:
...
       leakDetectionThreshold: 60000
+
   jpa:
     properties:
       hibernate:
         format_sql: true
     hibernate:
       ddl-auto: none
-    show-sql: true
+    show-sql: false
```
Kartiiyer12, line 22: Hi @matiwinnetou Is there a reason why this is turned off? This helps in easier debugging.
Also, for all the log that accumulates over time, do we need have auto pruning after a certain period of time. If yes, something to be looked after in separate issue.

PR #321, opened 2025-03-11
feat: exposing configuration variables for yaci-store prunning inside of rosetta yaci indexer

### LINKED ISSUE #294: Add pruning-safe-blocks configuration for yaco-store pruning.
## Why
There can be use cases where users may want to check transactions till a particular time since it happened. For this, they will have to disable pruning feature. 

## What
There is another configuration store.utxo.pruning-safe-blocks that can be configured in the application properties which can be used to specify the number of safe blocks to keep in the store. By default, this is set to 2160 blocks, same as the security parameter in Cardano mainnet. 
Adding an env variable to set this property will avoid deleting the spent utxos for that particular no. of blocks for users who want to preserve the blocks for say 24 hours.

Also, add proper documentation for env variable for user understanding.

matiwinnetou:
Fixes #294 

Adding PRUNING_SAFE_BLOCKS and PRUNING_INTERVAL for Yaci-store pruning.


Kammerlo (approved):

tienna (approved):
I tested adding these new variables on the Preview network. variables work fine

matiwinnetou:
@tienna do we want to test on the main-net instance or are we comfortable to merge on your side and put this into 1.2.4 release?

matiwinnetou:
Some exchanges have already implemented similar approach. We tested it on preprod and we believe the variable would be properly propagated on main-net.

PR #341, opened 2025-03-13
fix: we apply application specific index via flyweel instead of bash script

matiwinnetou:
Fixes #362


matiwinnetou:
I verified this on preprod machine upon reindexing from scratch:

image

PR #366, opened 2025-03-25
feat: initialize documentation structure and deployment workflow

linconvidal:


matiwinnetou (approved):
Are we able to make documentation specific to a given Rosetta version?

linconvidal:
@matiwinnetou [Docusaurus has support for that,](https://docusaurus.io/docs/versioning) just need to set it up.

PR #368, opened 2025-03-25
chore: JDK24 LTS

matiwinnetou:


matiwinnetou:
Blocked by Yaci-Indexer on Spring Boot 3.4 and https://github.com/bloxbean/yaci-store/issues/539

PR #372, opened 2025-03-26
docs: add an example script to sign rosetta transactionadd an example script to sign rosetta transaction

tienna:
This PR will add:
- [x] example script written in CSL to sign Rosetta transactions
- [x] short user guide and examples


matiwinnetou:
Let's wait for it until we tidy up our documentation and move to the new documentation system.

PR #389, opened 2025-04-02
feat: enable JDK 21 virtual threads for Spring Boot itself

matiwinnetou:


matiwinnetou:
We are fully loaded now but we may integrate this for 1.2.7 or later

PR #390, opened 2025-04-02
chore: cardano-node 10.2.1 integrated.

matiwinnetou:


api/src/test/resources/config/application-test-integration.yaml
Kartiiyer12, line 1: @matiwinnetou Should this not be 10.2.1 instead of 10.1.2?

PR #402, opened 2025-04-09
fix: improve performance on /account/balance for addresses with large transactions count. (quick wins)

matiwinnetou:


linconvidal:
Waiting for client confirmation of need

Issue #377, opened 2025-04-01
Docker compose option to build all the images from the source


Kartiiyer12:
One of the requirements for rosetta is to have the ability to build images from the source rather than using the pre-built option. This is the reason, one DockerFile image is adopted so that exchanges can easily build from source.

Its better if we provide option to build from source all the images in docker compose setup. 

Currently a few images are not build from source:
- cardano-submit-api
- cardano-node (albeit Docker file is there)
-  postgres

All software should build via docker in docker-compose setup as well.

VladislavKudrin:
#420 #438 #442

PR #420, opened 2025-04-21
feat: docker compose build from source

VladislavKudrin:
#377


docker/dockerfiles/postgres/Dockerfile
```diff
@@ -0,0 +1,2 @@
+ARG DB_IMAGE_TAG
+FROM postgres:${DB_IMAGE_TAG}
```
matiwinnetou, line 2: no, we actually need to download postgres at a particular version (using git clone) and build... it via make

matiwinnetou (changes requested):
postgres must be build via make and sources downloaded.... via git clone

matiwinnetou (approved):

PR #460, opened 2025-05-20
fix: cardano node db path / mithril version / tags

VladislavKudrin:


.env.docker-compose
```diff
@@ -32,6 +32,7 @@ CARDANO_NODE_DB=${CARDANO_NODE_DIR}/db
 CARDANO_CONFIG=./config/${NETWORK}
 
 ## Mithril
+MITHRIL_VERSION=2517.1
```
matiwinnetou, line 35: why is MITHRIL_VERSION twice in this file?

.env.docker-compose-preprod
```diff
@@ -33,6 +33,7 @@ CARDANO_NODE_DB=${CARDANO_NODE_DIR}/db
 CARDANO_CONFIG=./config/${NETWORK}
 
 ## Mithril
+MITHRIL_VERSION=2517.1
```
matiwinnetou, line 36: why is MITHRIL_VERSION twice in this file?

matiwinnetou (changes requested):
Mithril version twice in the env files?

PR #464, opened 2025-05-26
feat: split node submit containers

VladislavKudrin:


docker/dockerfiles/node/entrypoint.sh
```diff
@@ -1,27 +1,32 @@
...
-cardano-node run --socket-path "$CARDANO_NODE_SOCKET_PATH" --port $CARDANO_NODE_PORT --database-path "$CARDANO_NODE_DB" --config /config/config.json --topology /config/topology.json &
-CARDANO_NODE_PID=$!
-sleep 2
-
-echo "Starting Cardano submit api..."
-cardano-submit-api --listen-address 0.0.0.0 --socket-path "$CARDANO_NODE_SOCKET_PATH" --port $NODE_SUBMIT_API_PORT $NETWORK_STR  --config /cardano-submit-api-config/cardano-submit-api.yaml &
-CARDANO_SUBMIT_PID=$!
-
-wait -n $CARDANO_NODE_PID $CARDANO_SUBMIT_PID
+cmd="$1"; shift
```
matiwinnetou, line 9: This is also possible but why not have 2 different entry points?

matiwinnetou (approved):

PR #479, opened 2025-06-05
fix: drep vote delegation account not needed

matiwinnetou:


matiwinnetou:
@linconvidal 

Thank you very much for the thoughtful analysis you presented. I really appreciate the clarity and effort you’ve put into it.

That said, I’m not yet as confident in reaching the same conclusion regarding whether we should keep or remove the field. 
Let me briefly explain my perspective.

I fully understand and respect your point that retaining the field may reduce user confusion, as it aligns with the standardised structure of balance-based operations. However, I see it a bit differently.
If we’re including a field that the code doesn’t actually use, it essentially becomes dead code. From a user’s perspective, one could enter arbitrary data like “blabla,” and the system wouldn’t flag it. As a developer, I then have to consider whether this data needs to be validated, cross-checked (e.g., deriving a public address and confirming it matches the Bech32 representation), or simply ignored—which adds complexity and ambiguity.

Let me know what you think. Perhaps I am missing something.

PR #502, opened 2025-06-20
feat: upgrade to yaci-store 0.1.4, yaci-core 0.3.7 which has store.continue-on-parse-error in case there is a problem processing a block.

matiwinnetou:


Kartiiyer12:
more details - https://github.com/cardano-foundation/cardano-rosetta-java/issues/506

PR #503, opened 2025-06-23
feat: prefix support drep

VladislavKudrin:


api/src/main/java/org/cardanofoundation/rosetta/api/block/model/domain/DRepDelegation.java
```diff
@@ -26,18 +31,50 @@ public static class DRep {
...
             DRepType dRepType = switch (dRepParams.getType()) {
                 case KEY_HASH -> DRepType.ADDR_KEYHASH;
                 case SCRIPT_HASH -> DRepType.SCRIPTHASH;
                 case ABSTAIN -> DRepType.ABSTAIN;
                 case NO_CONFIDENCE -> DRepType.NO_CONFIDENCE;
             };
 
-            return new DRepDelegation.DRep(dRepParams.getId(), dRepType);
+            if (dRepType == DRepType.ABSTAIN || dRepType == DRepType.NO_CONFIDENCE) {
+                return new DRep(null, dRepType);
```
matiwinnetou, line 47: Are you sure this is correct that id is in such case null?

to discuss...

```
    public static DRep toDrep(String drepId, DRepType drepType) {
        byte[] bytes = Bech32.decode(drepId).data;

        if (drepType == DRepType.ADDR_KEYHASH) {
            return DRep.addrKeyHash(HexUtil.encodeHexString(bytes));
        } else if (drepType == DRepType.SCRIPTHASH) {
            return DRep.scriptHash(HexUtil.encodeHexString(bytes));
        } else {
            throw new IllegalArgumentException("Invalid DrepType");
        }
    }
```

api/src/test/java/org/cardanofoundation/rosetta/api/construction/service/PayloadsApiTest.java
```diff
@@ -120,4 +137,17 @@ private void assertConstructionPayloads(String requestFile, String expectedUnsig
...
+    ConstructionPayloadsRequest request = getPayloadRequest(requestFile);
+
+    ApiException exception = Assertions.assertThrows(
+            ApiException.class,
+            () -> constructionApiService.constructionPayloadsService(request)
+    );
+
+    System.out.println(exception.getMessage());
+
+    assertEquals(expectedErrorMessage, exception.getMessage());
```
matiwinnetou, line 150: I would rather verify error code because verifiy message error is prone to translation errors (I had this on one of my machines) for different locale.

PR #526, opened 2025-07-06
chore: bump mithril convert snapshot / legacy, lmdb snapshots support

VladislavKudrin:


matiwinnetou:
Confusing, let's discuss this.

PR #563, opened 2025-08-19
Fix: Complete search transaction results by including tx_input transactions

Kartiiyer12:
**Summary**

  Enhanced the search transactions API to return complete transaction history by including transactions where addresses spent UTXOs (input transactions) in addition to transactions
  where addresses received UTXOs (output transactions).

  **Problem**

  The search transactions API (/search/transactions) was only returning partial results. It only found transactions where the queried address received outputs (address_utxo table) but
  missed transactions where the address spent inputs without receiving outputs back to the same address (tx_input table).

  For example, address addr_test1wqur0rcryxwsgjccpfavct0ughwhje85s3cy853wxgphwkg2rshys was returning 109 transactions instead of the complete 163 transactions.

  **Solution**

  - Added new repository method: findCompleteTransactionHistoryByAddress() in AddressUtxoRepository
  - Enhanced SQL query: Uses UNION to combine both output and input transactions:
    - Output transactions: Where address received UTXOs (address_utxo table)
    - Input transactions: Where address spent UTXOs (tx_input joined with address_utxo)
  - Updated service: Modified LedgerSearchServiceImpl to use the complete method instead of the output-only method


Kartiiyer12:
There is also another fix that was done where the transaction.update_datetime.desc was added for sorting. This will fix the sorting of transactions  with-in block. 
This works in case-by-case basis. There are txns which has difference in milliseconds and some do not. 

  Performance Comparison by Address Size:

  | Address Size | WITHOUT UPDATE_DATETIME.desc() | WITH UPDATE_DATETIME.desc() | Performance Impact |
  |--------------|--------------------------------|-----------------------------|--------------------|
  | 2,770 txs    | 0.28s basic                    | 0.28s basic                 | No impact          |
  | 10,530 txs   | 1.00s basic                    | 1.41s basic                 | +41% slower        |
  | 15,682 txs   | 1.18s basic                    | 1.21s basic                 | +3% slower         |
  | 119,207 txs  | 7.95s basic                    | 8.42s basic                 | +6% slower         |

PR #591, opened 2025-09-09
fix: increase pruning interval

Kartiiyer12:
- 10 minute interval for pruning is too aggressive and it causes very high I/O when syncing. Increasing it to 1 hour.
    Rationale - 
     -  Constant insert/delete cycles on address_utxo table
     - Index Maintenance: Heavy I/O from index rebuilds after 264M deletions
     - Write-Heavy Workload: 15:1 write-to-read ratio overwhelming storage
- configure option for pruning batch size. Default is 3000 but if needed can be changed  upto 10000 for faster deletion but it should not cause too much I/O while syncing
 
- debug option for utxo store


.env.h2
```diff
@@ -42,6 +42,8 @@ SEARCH_LIMIT=100
 INDEXER_DOCKER_IMAGE_TAG=main
 REMOVE_SPENT_UTXOS=false
 REMOVE_SPENT_UTXOS_LAST_BLOCKS_GRACE_COUNT=2160
+REMOVE_SPENT_UTXOS_BATCH_SIZE:3000
```
matiwinnetou, line 45: this is incorrect, should be equals

PR #734, opened 2026-04-10
docs: clarify Kubernetes startup and readiness flow

ghost:
- document node sync and schema gates in Helm/Kubernetes startup
  - clarify API readiness, LIVE transition, and index-applier behavior
  - note current Helm offline-mode limitations and related values


Sotatek-DucPhung:
@linconvidal I’ve updated the docs for consistency with the current Helm and container flow. Could you take a look?
Thanks!

PR #746, opened 2026-04-20
feat: replace 'apply indexes' with a native solution

### LINKED ISSUE #736: Spike: replace 'apply indexes' with a native solution
TBD

ghost:
# feat: replace 'apply indexes' with a native solution

Closes #736.

Moves the required Rosetta index lifecycle (create, validate, track progress)
from the external `index-applier` sidecar container into `yaci-indexer` itself.
After blockchain sync reaches tip, the indexer automatically creates all 16
Rosetta-required indexes, gates readiness on completion, and exposes per-index
lifecycle state via a new actuator endpoint.

---

## Architecture: Before vs After

```
BEFORE
──────────────────────────────────────────────────────────────
  ┌─────────────────┐     ┌──────────────────┐
  │  yaci-indexer   │     │  index-applier   │
  │                 │     │  (sidecar)       │
  │  sync chain     │     │                  │
  │  expose data    │     │  apply-indexes   │
  │                 │     │  .sh             │
  └─────────────────┘     └──────────────────┘
          │                        │
          │                        │  psql + shell loop
          └──────────┬─────────────┘
                     ▼
              PostgreSQL DB
              (16 indexes)

AFTER
──────────────────────────────────────────────────────────────
  ┌───────────────────────────────────────────┐
  │              yaci-indexer                 │
  │                                           │
  │  sync chain                               │
  │  expose data                              │
  │  PgIndexService                           │
  │    └─ creates indexes when synced         │
  │    └─ gates readiness probe               │
  │    └─ exposes /actuator/rosetta-indexes   │
  └───────────────────────────────────────────┘
                     │
                     │  JDBC (autoCommit=true)
                     ▼
              PostgreSQL DB
              (16 indexes)
```

---

## Lifecycle State Machine

```mermaid
stateDiagram-v2
    [*] --> PENDING : startup (init)

    PENDING --> PENDING : sync not yet at tip
    PENDING --> APPLYING : node reaches tip\n(compareAndSet guard)

    APPLYING --> READY : all 16 indexes created\nsuccessfully
    APPLYING --> FAILED : ≥1 index failed\n(remaining still attempted)

    READY --> [*]
    FAILED --> [*]

    note right of PENDING
        Re-queries pg_index on startup.
        Skips to READY if all indexes
        already exist and are valid.
    end note

    note right of APPLYING
        Single background thread.
        Continue-on-failure: all 16
        indexes are always attempted.
    end note

    note right of FAILED
        Per-index errorMessage available
        via /actuator/rosetta-indexes.
        Recovery requires pod restart.
    end note
```

---

## Index Creation Flow

```mermaid
flowchart TD
    A([triggerIndexing called]) --> B{state ==\nPENDING?}
    B -- No --> Z([return — no-op])
    B -- Yes --> C[CAS: PENDING → APPLYING]
    C --> D[submit to RosettaIndexBuilder thread]

    D --> E[for each of 16 indexes]

    E --> F[queryIndexState from pg_index\nschema-filtered]

    F --> G{state?}

    G -- READY --> N[skip — already done]
    G -- INVALID or BUILDING --> H[DROP INDEX IF EXISTS\nschema.indexname]
    G -- MISSING --> I[CREATE INDEX CONCURRENTLY\nautoCommit=true]
    H --> I

    I --> J{success?}
    J -- Yes --> K[mark READY\nupdate lastProgressAt]
    J -- No --> L[mark FAILED\nlog error\nfailedCount++]

    K --> M{more indexes?}
    L --> M

    N --> M
    M -- Yes --> E
    M -- No --> O{failedCount == 0?}

    O -- Yes --> P[state = READY]
    O -- No --> Q[state = FAILED\nlog warning]
```

---

## Health Probe Behaviour

```mermaid
flowchart LR
    subgraph Readiness [Readiness Probe — /actuator/health/readiness]
        R1{connection\nalive?} -- No --> RD1[DOWN\nConnection]
        R1 -- Yes --> R2{synced\nto tip?}
        R2 -- No --> RD2[DOWN\nSyncing]
        R2 -- Yes --> R3{index lifecycle\n== READY?}
        R3 -- No --> RD3[DOWN\nApplying Indexes]
        R3 -- Yes --> RU[UP]
    end

    subgraph Liveness [Liveness Probe — /actuator/health/liveness]
        L1{state ==\nAPPLYING?} -- No --> LU[UP]
        L1 -- Yes --> L2{lastProgressAt\n< 15 min ago?}
        L2 -- Yes --> LU
        L2 -- No --> LD[DOWN\nStalled]
    end
```

> **Note**: FAILED state → Readiness DOWN, Liveness **UP**. This is intentional —
> a stalled build warrants a pod restart (liveness kill), but a permanently failed
> index does not. Operators diagnose via `/actuator/rosetta-indexes`.

---

## Actuator Endpoint

`GET /actuator/rosetta-indexes`

```json
{
  "overallState": "APPLYING",
  "lastProgressAt": "2026-04-22T10:31:05Z",
  "totalRequired": 16,
  "totalReady": 9,
  "totalMissing": 6,
  "totalFailed": 1,
  "indexes": [
    { "name": "idx_address_utxo_amounts_gin", "state": "READY",   "errorMessage": null },
    { "name": "idx_tx_input_tx_hash",         "state": "BUILDING", "errorMessage": null },
    { "name": "idx_block_hash",               "state": "FAILED",   "errorMessage": "permission denied for table block" },
    { "name": "idx_stake_address_hash_raw",   "state": "MISSING",  "errorMessage": null }
  ]
}
```

---

### Why `autoCommit=true`?
`CREATE INDEX CONCURRENTLY` is rejected by PostgreSQL inside a transaction block.
The service uses a raw JDBC connection with explicit `autoCommit=true` to guarantee
no transaction wrapping.

### Why continue-on-failure?
The 16 indexes span independent tables. All indexes are always attempted so a single
permission error on one table does not block the remaining 15. Per-index failure
detail is available via `/actuator/rosetta-indexes`.

### Schema-qualified DROP
Both `pg_index` queries and `DROP INDEX` statements are scoped to `current_schema()`
to avoid cross-schema matches in multi-schema PostgreSQL deployments.

---

## Infrastructure Changes

| Area | Before | After |
|------|--------|-------|
| Docker Compose | `index-applier` sidecar container | Removed — `yaci-indexer` handles indexes natively |
| `apply-indexes.sh` | Shell script (psql + yq loop) | Deleted |
| Postgres Dockerfile | Bundled `jq`, `yq`, `apply-indexes.sh` | Removed (no longer needed) |
| Helm chart | `index-applier` Job + ConfigMap templates | Removed — `indexApplier` section dropped |
| Helm profile resources | `indexApplier` blocks in entry/mid/advanced | Removed |
| Healthcheck retries | 360 × 10 s = 1 hour | 1080 × 10 s = **3 hours** (mainnet GIN build time) |
| Documentation | References to `index-applier` throughout | Updated to reflect native approach |

---

## Package & Class Naming (yaci-indexer)

Classes were renamed for clarity as part of this PR:

| Old | New |
|-----|-----|
| `indexmanagement` package | `indexes` package |
| `RosettaIndexLifecycleService` | `IndexService` |
| `PostgreSQLRosettaIndexLifecycleService` | `PgIndexService` |
| `NoOpRosettaIndexLifecycleService` | `NoOpIndexService` |
| `RosettaIndexConfig` | `IndexCatalog` |
| `RosettaIndexesEndpoint` | `IndexesEndpoint` |
| `RosettaIndexProgressSnapshot` | `IndexProgress` |
| `RosettaIndexStallIndicator` | `IndexStallIndicator` |

---

## API module — cleanup

Removed dead `getIndexCommands()` from `RosettaIndexConfig` in the api module.
The api never creates indexes — that is `PgIndexService`'s responsibility.
The method was unused and has been deleted.

---

## Test Coverage

39 unit tests across 5 classes:

| Class | Scenarios |
|-------|-----------|
| `PgIndexServiceTest` | Init states, sync trigger, INVALID drop, BUILDING drop, continue-on-failure, failure handling, double-trigger guard |
| `IndexesEndpointTest` | APPLYING snapshot, READY (all pre-exist) snapshot |
| `IndexStallIndicatorTest` | Stall detection, terminal state distinction |
| `YaciSyncHealthIndicatorTest` | Readiness states including APPLYING_INDEXES |
| `IndexCatalogParityTest` | Both module YAML files are byte-identical |

> `CREATE INDEX CONCURRENTLY` with `autoCommit=true` must be verified against a real
> PostgreSQL instance — unit mocks cannot catch transaction-block rejections.


docs/docs/advanced-configuration/index-management.md
```diff
@@ -8,38 +8,74 @@ description: Understanding and customizing database index creation
 
 ## Overview
 
-Starting from v2.0, Rosetta uses deferred index creation to improve initial deployment time. The `index-applier` container automatically creates required indexes after initial sync completes.
+Starting from v2.1, Rosetta manages the required database indexes natively inside `yaci-indexer`. After blockchain sync reaches tip, the indexer automatically creates all required indexes without any external sidecar container.
```
matiwinnetou: Version is wrong, it will be from V2.3.0 as things stand at the moment.

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/health/IndexStallIndicator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.time.Instant;
+
+@Component("indexStall")
+public class IndexStallIndicator implements HealthIndicator {
+
+    private final IndexService indexService;
+    private final int stallTimeoutMinutes;
+
+    public IndexStallIndicator(IndexService indexService,
+                                      @Value("${cardano.rosetta.index.stall-timeout-minutes:15}") int stallTimeoutMinutes) {
```
matiwinnetou: Cosmetic (optional) alternative:
```
@RequiredArgsConstructor
public class IndexStallIndicator {

    private final IndexService indexService;

    @Value("${cardano.rosetta.index.stall-timeout-minutes:15}")
    private int stallTimeoutMinutes;
}
```

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/indexes/PgIndexService.java
```diff
@@ -0,0 +1,263 @@
...
+
+    @Override
+    public List<IndexItemStatus> getIndexStatus() {
+        if (indexConfig.getDbIndexes() == null || indexConfig.getDbIndexes().isEmpty()) {
+            return new ArrayList<>(itemStatusMap.values());
+        }
+
+        List<IndexItemStatus> liveStatuses = new ArrayList<>();
+        for (IndexCatalog.DbIndex dbIndex : indexConfig.getDbIndexes()) {
+            IndexItemState liveState = queryIndexState(dbIndex.name());
```
matiwinnetou, line 137: **[HIGH — bug]** `getIndexStatus()` re-queries `pg_index` and always uses the live state, which can only ever be `MISSING / INVALID / BUILDING / READY` — the in-memory `FAILED` state is always overwritten (only `errorMessage` survives). As a result `/actuator/rosetta-indexes` can never report `"state": "FAILED"` and `totalFailed` is always `0`, contradicting the PR description and the `index-management.md` example. An operator can see `overallState: FAILED` while every item shows `MISSING`/`INVALID`, which is confusing.

Suggestion: when the cached entry is `FAILED` and the live state isn't `READY`, keep `FAILED` (or derive it from `errorMessage != null`).

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/indexes/PgIndexService.java
```diff
@@ -0,0 +1,263 @@
...
+
+    @Override
+    public Instant getLastProgressAt() {
+        return lastProgressAt.get();
+    }
+
+    @Scheduled(fixedDelay = 30000)
+    public void checkSyncAndTrigger() {
+        IndexLifecycleState currentState = state.get();
+        if (currentState != IndexLifecycleState.PENDING && currentState != IndexLifecycleState.FAILED) {
```
matiwinnetou, line 164: **[HIGH — design / correctness of retry behaviour]** This re-triggers from `FAILED` every 30 s with no backoff. For a *persistent* failure this loops forever — the disk-full case is the bad one: each cycle drops the leftover invalid index and re-runs a potentially hours-long `CREATE INDEX CONCURRENTLY`, indefinitely churning the DB. Liveness never fires because `triggerIndexing()` resets `lastProgressAt` on every retry.

Suggestions:
1. Add a retry cap or exponential backoff for the `FAILED` → `APPLYING` transition.
2. The PR description's state diagram (`FAILED --> [*]`, "Recovery requires pod restart") and the `IndexLifecycleState.FAILED` comment ("operator action required") predate this retry behaviour — please update whichever semantics we keep.

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/indexes/PgIndexService.java
```diff
@@ -0,0 +1,263 @@
...
+                // Re-query pg_index per index so that any future retry path (e.g. manual re-trigger)
+                // always detects and drops invalid/building rows before re-attempting — never rely on
+                // cached state from a previous run.
+                IndexItemState currentItemState = queryIndexState(dbIndex.name());
+
+                // Treat BUILDING (from prior JVM crash) the same as INVALID — drop and rebuild.
+                // Schema-qualified drop (Tier 1.3) avoids matching wrong index in multi-schema DBs.
+                if (currentItemState == IndexItemState.INVALID || currentItemState == IndexItemState.BUILDING) {
+                    log.info("Index {} is {}. Dropping before rebuild.", dbIndex.name(), currentItemState);
+                    jdbcTemplate.execute("DROP INDEX IF EXISTS " + currentSchema + "." + dbIndex.name());
```
matiwinnetou: **[MEDIUM — operational risk]** Plain `DROP INDEX` takes an `ACCESS EXCLUSIVE` lock on the table and queues behind (and then blocks) running queries — and at this point the api module may already be serving traffic against the same tables.

Suggestion: use `DROP INDEX CONCURRENTLY IF EXISTS`, executed via `executeWithAutoCommit()` — like `CREATE … CONCURRENTLY` it cannot run inside a transaction block, so it should go through the raw autocommit connection rather than `jdbcTemplate`.

matiwinnetou (commented):
Follow-up review pass on the index lifecycle implementation. Each inline comment is tagged with a priority: **HIGH** (should be addressed before merge), **MEDIUM** (operational risk, strongly recommended), **LOW / cosmetic** (optional, can be deferred or done as follow-up).

Two additional notes that don't anchor well to a diff line:

- **LOW / informational**: running two `yaci-indexer` replicas would race on `CREATE INDEX CONCURRENTLY` — `IF NOT EXISTS` makes one replica silently "succeed" while the other is still building. Fine under the single-replica assumption, but worth a one-line note in the Helm values.
- **Cosmetic, pre-existing (not from this PR)**: `yaci-indexer/src/main/resources/application.properties` has a stray `7` on its own line (~line 42) and a trailing `a` in a comment banner (~line 40) — easy drive-by cleanup if you touch the file again.

What looks good: schema-scoped `pg_index` queries, treating leftover `BUILDING` as droppable after a crash, the raw autocommit connection for `CONCURRENTLY`, continue-on-failure across independent tables, and the `IndexCatalogParityTest` guard.

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/indexes/PgIndexService.java
```diff
@@ -0,0 +1,263 @@
...
+            urlWithKeepAlive += urlWithKeepAlive.contains("?") ? "&tcpKeepAlive=true" : "?tcpKeepAlive=true";
+        }
+        return DriverManager.getConnection(urlWithKeepAlive, dbUser, dbPassword);
+    }
+
+    private void executeWithAutoCommit(String sql) {
+        try (Connection conn = getConnection()) {
+            conn.setAutoCommit(true);
+            try (Statement stmt = conn.createStatement()) {
+                stmt.setQueryTimeout(21600); // 6-hour timeout for long-running index creations
```
matiwinnetou: **[MEDIUM — config / resilience]** This hardcoded 6 h query timeout exactly equals the default `INDEX_STALL_TIMEOUT_MINUTES=360`, and `lastProgressAt` only advances *between* indexes. If a single index build (e.g. the mainnet GIN) exceeds 6 h: liveness goes DOWN → pod killed mid-build → leftover `BUILDING` index dropped on restart → full rebuild → potential restart loop.

Suggestions:
1. Make this timeout configurable alongside `INDEX_STALL_TIMEOUT_MINUTES` rather than hardcoded.
2. Document the invariant: stall timeout must exceed the longest single index build.
3. Nit: the PR body still says liveness stall is "15 min" — the effective default is 360 via `application.properties`; the `:15` fallback in `IndexStallIndicator` is dead in production.

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/indexes/IndexesEndpoint.java
```diff
@@ -0,0 +1,43 @@
+package org.cardanofoundation.rosetta.yaciindexer.indexes;
+
+import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.stereotype.Component;
+
+import lombok.RequiredArgsConstructor;
+
+@Component
+@Endpoint(id = "rosetta-indexes")
```
matiwinnetou: **[LOW — informational, optional]** Spring Boot logs an "endpoint ID contains invalid characters" warning at startup for hyphenated endpoint ids. It does work as-is — exposure matching normalizes `rosettaIndexes` ↔ `rosetta-indexes` (both reduce to the same alphanumeric form), so the endpoint is exposed and reachable at the documented URL.

If we want to get rid of the warning *without* changing the URL, the id and the URL path can be decoupled via `management.endpoints.web.path-mapping`:

```java
@Endpoint(id = "rosettaIndexes")   // valid id → no startup warning
```

```properties
management.endpoints.web.path-mapping.rosettaIndexes=rosetta-indexes
```

Result:
- URL stays `/actuator/rosetta-indexes` — docs, PR description, and any operator tooling are unaffected.
- The existing exposure entry `management.endpoints.web.exposure.include=...,rosettaIndexes` becomes an exact id match instead of relying on normalization.
- No "invalid characters" warning in the logs.

Still optional — current code is functional; this just makes the id conform to Spring's naming rules while keeping the nicer kebab-case URL.

yaci-indexer/src/main/java/org/cardanofoundation/rosetta/yaciindexer/health/YaciSyncHealthIndicator.java
```diff
@@ -74,6 +87,15 @@ public Health health() {
                     .build();
         }
 
+        IndexLifecycleState indexState = indexService.getState();
+        builder.withDetail("indexLifecycleState", indexState.name());
+
+        if (indexState != IndexLifecycleState.READY) {
+            return builder.down()
+                    .withDetail("syncStatus", "Applying Indexes")
```
matiwinnetou: **[LOW — cosmetic, optional]** This `indexState != READY` branch catches 3 of the 4 lifecycle states — `PENDING`, `APPLYING`, and `FAILED` — and emits the identical `syncStatus: "Applying Indexes"` for all of them. For two of those the message is inaccurate:

- **`PENDING`** — node is synced but the build hasn't been triggered yet (the 30 s scheduler simply hasn't fired): nothing is being applied at that moment.
- **`FAILED`** — a build round ended with failures (`PgIndexService` sets `FAILED` at the end of a run). With the 30 s retry the state oscillates `FAILED → APPLYING → FAILED`, so readiness intermittently reports `indexLifecycleState: FAILED` while the human-readable message still claims indexes are being applied.

Nothing is functionally wrong — the `indexLifecycleState` detail on the line above carries the exact state, and the probe is correctly DOWN in all three cases. This is purely about the operator-facing `syncStatus` string.

Alternative that could be taken — map each state explicitly; probe semantics stay identical (DOWN for everything non-READY), only the message improves:

```java
if (indexState != IndexLifecycleState.READY) {
    String syncStatus = switch (indexState) {
        case PENDING  -> "Awaiting index creation";
        case APPLYING -> "Applying Indexes";
        case FAILED   -> "Index creation failed — see /actuator/rosetta-indexes";
        case READY    -> throw new IllegalStateException("unreachable");
    };
    return builder.down()
            .withDetail("syncStatus", syncStatus)
            .build();
}
```

(Or, simpler: keep the single string but rename it to something state-neutral like `"Indexes not ready"`.)

PR #758, opened 2026-06-15
fix: block data endpoints when indexer is syncing

Sotatek-HenryDo:
Ref: #720 

Blocked Rosetta data endpoints from responding when the indexer is syncing, requiring the sync stage to be "LIVE" in order to return data.

Introduced a retriable early-rejection check returning a newly defined dedicated Rosetta Error Code: 5056 (INDEXER_NOT_READY) with a clear explanation message on readiness.


Kartiiyer12:
- Ordering nit (account/block/search): verifySyncStatus() is now called before verifyNetworkRequest() (e.g.
  AccountApiImplementation.java:37-38, BlockApiImpl.java:42, SearchApiImpl.java:42-44). That means a request with an invalid network_identifier while the indexer is syncing returns 5056 INDEXER_NOT_READY (retriable=true) instead of invalidBlockchainError/networkNotFoundError. Clients will retry forever instead of fixing their request.
  Swap the order so that first check is for network and then the sync check.  
  
  - The network status check,  hit the db for every request That would mean an extra request for every request. Once the stage is LIVE, we dont expect this to go down quickly. So maybe once its live, we can cache and refresh only 5~10 secs to avoid constant checking for every request. There needs to be a proper performance load testing for this to ensure do not impact the performance.

Issue #759, opened 2026-06-18
cache network status check after live


Sotatek-HenryDo:
Currently the data endpoints check the indexer sync status before serving requests.

This hits the db on every request. Once the stage is LIVE, we dont expect it to go down quickly, so we could cache the LIVE state and refresh it every 5-10 seconds instead of checking for every request.

There should be proper performance/load testing before doing this, to make sure the network status check does not impact data endpoint performance

PR #761, opened 2026-06-26
feat: add 5s cache for sync status verification

Sotatek-HenryDo:
Ref: #759 

Introduced a short-lived (5-second by default, configurable) cache for the indexer sync status verification.

Once the indexer transitions to the "LIVE" stage, subsequent status verifications (`verifySyncStatus()`) utilize the cached status instead of querying the database. This significantly reduces database load and prevents repetitive checks on high-frequency endpoints (such as `/block` and `/account/balance`) under heavy load.

### DIFF
--- a/api/src/main/java/org/cardanofoundation/rosetta/api/network/service/SyncStatusHealthIndicator.java
+++ b/api/src/main/java/org/cardanofoundation/rosetta/api/network/service/SyncStatusHealthIndicator.java
@@ -1,7 +1,9 @@
 package org.cardanofoundation.rosetta.api.network.service;
 
 import lombok.RequiredArgsConstructor;
-import org.cardanofoundation.rosetta.api.block.service.LedgerBlockService;
+import java.util.Optional;
+import org.openapitools.client.model.SyncStatus;
+
 import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/cardanofoundation/rosetta/api/network/service/SyncStatusHealthIndicator.java
```diff
@@ -35,13 +37,19 @@
...
 
     @Override
     @Nullable
     public Health health() {
-        var latestBlock = ledgerBlockService.findLatestBlockIdentifier();
-        var syncStatusOpt = syncStatusService.calculateSyncStatus(latestBlock);
+        Optional<SyncStatus> syncStatusOpt;
+        try {
+            syncStatusOpt = syncStatusService.getSyncStatus();
+        } catch (Exception e) {
```
matiwinnetou: why are you try catching this here?

matiwinnetou: For which scenarios are we catching exception?

Sotatek-HenryDo: the only real case is findLatestBlockIdentifier() throwing genesisBlockNotFound when the indexer has no blocks yet

Sotatek-HenryDo: I'll drop the try/catch and have getSyncStatus() return Optional.empty() in that case so it falls into the existing "no block data" branch

PR #764, opened 2026-07-07
spike: enable n2c connection pooling in yaci-indexer

matiwinnetou:
## What

Enables yaci-store's **N2C (Node-to-Client) connection pool** in the yaci-indexer. The pool code already ships in the `yaci-store` dependency (`2.0.1`) but is **defaulted off** upstream — this just turns it on with fixed, tuned values.

Change is config-only, in `yaci-indexer/src/main/resources/application.properties`:

```properties
store.cardano.n2c-pool-enabled=true
store.cardano.n2c-pool-max-total=10
store.cardano.n2c-pool-min-idle=2
store.cardano.n2c-pool-max-idle=5
store.cardano.n2c-pool-max-wait-in-millis=10000
```

## Why

The yaci-indexer talks to the local node over **two independent protocols**:

- **N2N** (`CARDANO_NODE_HOST`/`PORT`) — the high-volume chain-sync/block-fetch stream that drives all block indexing. **Not touched by this change.**
- **N2C** (local socket, `n2c-socket`/`n2c-socat` profiles) — used for three on-demand paths:
  1. **Stake-account reward lookups** — `/account/balance` for a **stake address** issues a live `DelegationsAndRewardAccounts` local-state query to the node on **every request** (`AccountServiceImpl.getAccountInfo`). Payment-address balances come from the indexed Postgres UTXO data and do not use N2C.
  2. **Protocol parameter fetching** (Construction API).
  3. **Transaction submission** when routed through the local node rather than the cardano-submit-api sidecar.

With pooling disabled (the current default), every N2C request opens a fresh socket, performs the N2C handshake, runs the query, then tears the connection down. For deployments serving exchanges/wallets that poll account balances, the stake-reward path repeats this connect/handshake/teardown cycle on **every stake-address balance request**, causing needless connection churn against the node socket. Pooling reuses warm connections and removes that churn.

## Decisions

- **Hardcoded, not env-exposed.** These are internal tuning constants; operators should not mis-tune them, so no new environment variables were added.
- **Correct binding key.** Uses `store.cardano.n2c-pool-max-total` (binds to `n2cPoolMaxTotal`). yaci-store's own sample config documents this as `n2c-max-total`, which binds to a non-existent field and is silently ignored — that spelling is intentionally avoided, and the config comment calls this out.
- The config comment links to the upstream docs: https://store.yaci.xyz/docs/v2/stores/configuration#n2c-connection-pool-configuration

## Scope / risk

- Pool is only instantiated when N2C is configured (`n2c-socket` / `n2c-socat` profiles); no-op otherwise.
- No Java changes, no API changes, no new dependencies.

### DIFF
--- a/yaci-indexer/src/main/resources/application.properties
+++ b/yaci-indexer/src/main/resources/application.properties
@@ -150,6 +150,27 @@ store.continue-on-parse-error=${CONTINUE_PARSING_ON_ERROR:false}
 
 jobs.peer-discovery.enabled=${PEER_DISCOVERY:false}
 
+#####################################################################################################
+# N2C (Node-to-Client) connection pool
+#
+# The n2c local-state-query connection to the local Cardano node is used for stake-account reward
+# lookups (/account/balance for stake addresses), protocol parameter fetching and tx submission.
+# With pooling disabled (yaci-store default), every such call opens a fresh socket + n2c handshake
+# and tears it down afterwards. Pooling reuses warm connections and removes that per-request churn.
+#
+# NOTE: the "max total" key is store.cardano.n2c-pool-max-total (binds to n2cPoolMaxTotal). yaci-store's
+# own sample config uses n2c-max-total, which does NOT bind - do not copy that name.
+# Only takes effect when n2c is configured (n2c-socket / n2c-socat profiles).
+#
+# Upstream documentation:
+# https://store.yaci.xyz/docs/v2/stores/configuration#n2c-connection-pool-configuration
+#####################################################################################################
+store.cardano.n2c-pool-enabled=true
+store.cardano.n2c-pool-max-total=10
+store.cardano.n2c-pool-min-idle=2
+store.cardano.n2c-pool-max-idle=5
+store.cardano.n2c-pool-max-wait-in-millis=10000
+
 # disable local state for n2c governance for now
 #When enabled, yaci-store will periodically query the Cardano node's  current governance state via n2c protocol: proposal status, drep distribution,...
 #Before the  governance-aggr module was introduced, this was the only way to get the above governance data: proposal status,, drep distribution (voting power), current committee state...


linconvidal:
My only concern here is the saturation behavior of the N2C pool.

With `store.cardano.n2c-pool-max-total=10` and `store.cardano.n2c-pool-max-wait-in-millis=10000`, any scenario with more than 10 concurrent N2C operations will start queueing behind the pool. In particular, if an exchange/wallet is polling `/account/balance` for stake addresses, those requests all go through live N2C local-state queries, which can be relatively slow compared to DB-backed balance lookups.

This is probably still much better than opening a fresh socket/handshake per request, but I think the main thing to validate is the backpressure behavior under realistic stake-address polling.

Not a blocker from my side, but I would like us to either test this with concurrent stake-address balance requests or document why `10` is the right limit for now.

linconvidal:
### 1. Transaction submission uses `cardano-submit-api`, not the Yaci N2C pool

Conclusion: `/construction/submit` builds an HTTP URL from `CARDANO_NODE_SUBMIT_HOST`, `NODE_SUBMIT_API_PORT`, and `SUBMIT_API_PATH`, then posts the signed CBOR transaction with `RestTemplate.postForEntity`. The submit path is therefore the node submit API path, not the Yaci N2C local-state-query pool.

Evidence:

- [`CardanoConstructionServiceImpl.java`, submit host and port config](https://github.com/cardano-foundation/cardano-rosetta-java/blob/67faae61289de0b63520d5b28ba7cd84f05ba90a/api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java#L80-L84)
- [`CardanoConstructionServiceImpl.java`, `submitTransaction`](https://github.com/cardano-foundation/cardano-rosetta-java/blob/67faae61289de0b63520d5b28ba7cd84f05ba90a/api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java#L498-L508)
- [`Constants.java`, submit API path and success code](https://github.com/cardano-foundation/cardano-rosetta-java/blob/67faae61289de0b63520d5b28ba7cd84f05ba90a/api/src/main/java/org/cardanofoundation/rosetta/common/util/Constants.java#L13-L15)

<details>
<summary>Relevant code</summary>

```java
@Value("${cardano.rosetta.NODE_SUBMIT_API_PORT}")
private int nodeSubmitApiPort;

@Value("${cardano.rosetta.CARDANO_NODE_SUBMIT_HOST}")
private String cardanoNodeSubmitHost;
```

```java
public String submitTransaction(String signedTransaction) throws ApiException {
  String submitURL = Constants.PROTOCOL + cardanoNodeSubmitHost + ":" + nodeSubmitApiPort
          + Constants.SUBMIT_API_PATH;

  ResponseEntity<String> exchange = restTemplate.postForEntity(
          submitURL,
          new HttpEntity<>(decodeHexString(signedTransaction), headers),
          String.class);
```

```java
public static final String SUBMIT_API_PATH = "/api/submit/tx";
public static final int SUCCESS_SUBMIT_TX_HTTP_CODE = 202;
```

</details>

---

### 2. Protocol parameters are DB/cache reads in the API path, but Yaci Store populates the source table through N2C

Conclusion: `ProtocolParamServiceImpl.findProtocolParameters()` is cached with `protocolParamsCache`. In online mode, the Rosetta API reads protocol params from `local_epoch_param` through `LocalProtocolParamsRepository.getLocalProtocolParams()`, falling back to indexed `epoch_param` data through `epochParamRepository.findLatestProtocolParams()`. So the API request path is not a live N2C local-state query.

Pool interaction: Yaci Store populates `local_epoch_param` in the background via N2C. It runs `CurrentProtocolParamsQuery`, saves through `LocalEpochParamsStorage.save()`, and can consume one slot from the same N2C pool as stake-reward lookups. The default periodic refresh is every 5 minutes (`store.epoch.n2c-protocol-param-fetching-interval-in-minutes:5`), plus era/epoch event triggers. This is not per Rosetta API request.

Evidence:

- [`ProtocolParamServiceImpl.java`, API cache and repository lookup](https://github.com/cardano-foundation/cardano-rosetta-java/blob/67faae61289de0b63520d5b28ba7cd84f05ba90a/api/src/main/java/org/cardanofoundation/rosetta/common/services/ProtocolParamServiceImpl.java#L52-L67)
- [`LocalProtocolParamsRepository.java`, API query for latest local params](https://github.com/cardano-foundation/cardano-rosetta-java/blob/67faae61289de0b63520d5b28ba7cd84f05ba90a/api/src/main/java/org/cardanofoundation/rosetta/api/block/model/repository/LocalProtocolParamsRepository.java#L14-L18)
- [`LocalEpochParamService.java`, Yaci N2C `CurrentProtocolParamsQuery` and save](https://github.com/bloxbean/yaci-store/blob/4e527689ce762ffb625170f4a8e51b4339158671/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/service/LocalEpochParamService.java#L109-L145)
- [`LocalEpochParamsStorageImpl.java`, save to local epoch params storage](https://github.com/bloxbean/yaci-store/blob/4e527689ce762ffb625170f4a8e51b4339158671/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/storage/impl/LocalEpochParamsStorageImpl.java#L18-L26)
- [`LocalClientProviderManager.java`, same pool borrow path](https://github.com/bloxbean/yaci-store/blob/4e527689ce762ffb625170f4a8e51b4339158671/components/core/src/main/java/com/bloxbean/cardano/yaci/store/core/service/local/LocalClientProviderManager.java#L65-L73)
- [`LocalEpochParamsScheduler.java`, default 5-minute scheduled fetch](https://github.com/bloxbean/yaci-store/blob/4e527689ce762ffb625170f4a8e51b4339158671/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/processor/LocalEpochParamsScheduler.java#L31-L42)
- [`LocalEpochParamService.java`, era/epoch event-triggered fetches](https://github.com/bloxbean/yaci-store/blob/4e527689ce762ffb625170f4a8e51b4339158671/stores/epoch/src/main/java/com/bloxbean/cardano/yaci/store/epoch/service/LocalEpochParamService.java#L59-L86)

<details>
<summary>Relevant code</summary>

```java
@Cacheable(value = "protocolParamsCache")
@Transactional(readOnly = true)
public ProtocolParams findProtocolParameters() {
  if (!offlineMode) {
    Optional<LocalProtocolParamsEntity> protocolParams = localProtocolParamsRepository.getLocalProtocolParams();

    if (protocolParams.isEmpty()) {
      ProtocolParamsEntity paramsEntity = epochParamRepository.findLatestProtocolParams();
      return mapperProtocolParams.mapProtocolParamsToEntity(paramsEntity);
    }

    return protocolParams.get().getProtocolParams();
  }
```

```java
Optional<LocalClientProvider> localClientProvider = localClientProviderManager.getLocalClientProvider();
var localStateQueryClient = localClientProvider.map(LocalClientProvider::getLocalStateQueryClient).orElse(null);
Mono<CurrentProtocolParamQueryResult> mono = localStateQueryClient.executeQuery(new CurrentProtocolParamsQuery(era));
...
localProtocolParamsStorage.save(epochParam);
```

```java
if (isPoolEnabled) {
    var provider = localClientProviderPool.borrowObject();
    return Optional.of(provider);
}
```

</details>

PR #785, opened 2026-09-14
feat: support CIP-113 smart wallet address derivation

### LINKED ISSUE #768: 1: `/construction/derive` — CIP-113 Smart Wallet Address Derivation
### Background

The `/construction/derive` endpoint derives a Cardano account address from a public key. Cardano Rosetta will add a Cardano-specific `"CIP-113"` address type for the exchange enterprise-deposit convention.

It derives a CIP-113 smart-wallet base address whose payment credential is the configured `programmableLogicBase` (PLB) script hash and whose stake credential is the Blake2b-224 hash of the supplied enterprise payment public key.

This allows an exchange to derive the exchange-controlled CIP-113 smart-wallet deposit address corresponding to an existing enterprise deposit public key. 

Phase 1 supports one explicitly configured PLB script hash per Rosetta network instance. It does not query the CIP-113 registry, identify whether a policy is programmable, discover deployments, or validate CIP-113 transaction semantics.

### Derivation

```text
Input:
    public_key.hex_bytes = exchange enterprise payment public key
    public_key.curve_type = edwards25519
    metadata.address_type = CIP-113

publicKeyBytes = hexDecode(public_key.hex_bytes)
userCredential = Blake2b-224(publicKeyBytes)

SmartWallet = BaseAddress(
    payment = ScriptCredential(configuredPLBScriptHash),
    stake   = KeyCredential(userCredential),
    network = request.network_identifier
)
```

Credential sizes:

```text
Ed25519 public key:  32 bytes = 64 hexadecimal characters
User key hash:       28 bytes = 56 hexadecimal characters
PLB script hash:     28 bytes = 56 hexadecimal characters
```

The PLB hash is used directly as a script credential. It must not be padded, represented as a public key, or hashed again. A separate `metadata.staking_credential` is neither needed nor accepted.

### Request and response example

For this example, the server is configured with PLB script hash:

```text
f2182b00a37bd746e20575c9af01ab31312213514cd31e872e0a2a3e
```

Request:

```json
{
  "network_identifier": {
    "blockchain": "cardano",
    "network": "preprod"
  },
  "public_key": {
    "hex_bytes": "1b400d60aaf34eaf6dcbab9bba46001a23497886cf11066f7846933d30e5ad3f",
    "curve_type": "edwards25519"
  },
  "metadata": {
    "address_type": "CIP-113"
  }
}
```
Response:

```json
{
  "account_identifier": {
    "address": "addr_test1zreps2cq5daaw3hzq46untcp4vcnzgsn29xdx8589c9z504mgrc6v3au3rqm66mn3kuwke340kfxga82tl7kh2nke8as48q6r3"
  }
}
```

### Error behavior

Providing `metadata.staking_credential` with `address_type: "CIP-113"` must be rejected rather than ignored. If a new dedicated error is retained:

```json
{
  "code": 4060,
  "message": "Staking credential not allowed with CIP-113 address type",
  "retriable": false,
  "details": {
    "message": "For the exchange enterprise-address convention, CIP-113 derives the user credential from the top-level public_key and uses the configured PLB script hash as the payment credential. Do not provide metadata.staking_credential."
  }
}
```

The new error must be registered centrally and advertised by `/network/options`. Reusing an existing suitable invalid-metadata error is also acceptable if the issue and tests use it consistently.

`address_type` remains case-sensitive. `"cip113"` returns the existing `INVALID_ADDRESS_TYPE` error rather than introducing a duplicate error code:

```json
{
  "code": 4016,
  "message": "Provided address type is invalid",
  "retriable": false
}
```

### PLB configuration

Define and validate a dedicated property. Suggested name:

```yaml
cardano:
  rosetta:
    cip113:
      programmable-logic-base-script-hash: "f2182b00a37bd746e20575c9af01ab31312213514cd31e872e0a2a3e"
```

The final name should follow repository configuration conventions.

- Decode the value as hexadecimal.
- Require exactly 28 decoded bytes.
- Reject empty, non-hexadecimal, short, and long values.
- Use it directly as a script hash; do not hash it again.
- Use one configured PLB per Rosetta network instance for Phase 1.
- Do not silently substitute a default deployment.
- If CIP-113 is enabled, fail startup for missing or malformed configuration. If support is optional, fail the CIP-113 request with a documented configuration error.

### Implementation approach

Add a dedicated derivation method because ordinary base derivation hashes two public keys, while CIP-113 combines a configured script hash with a key-derived credential:

```java
String getCip113Address(PublicKey userPublicKey, NetworkEnum network);
```

Reuse the Cardano Client Library through `CardanoAddressUtils` for the address header, network tag, and Bech32 encoding:

```java
public String getCip113Address(PublicKey userPublicKey, NetworkEnum network) {
    byte[] userKeyHash = getHdPublicKeyFromRosettaKey(userPublicKey).getKeyHash();
    byte[] plbScriptHash = cip113Config.getPlbScriptHashBytes();

    return CardanoAddressUtils.getAddress(
            plbScriptHash,
            userKeyHash,
            (byte) 0x10,
            network.getNetwork(),
            com.bloxbean.cardano.client.address.AddressType.Base)
        .getAddress();
}
```

Prefer a named constant for the script-payment/key-stake address kind instead of an unexplained `0x10` literal.

### Acceptance criteria

- [ ] `POST /construction/derive` with `address_type: "CIP-113"` derives an exchange-controlled smart-wallet address from the supplied enterprise payment public key.
- [ ] The supplied public-key bytes are hashed with Blake2b-224.
- [ ] The result is a Cardano base address with a script payment credential and key stake credential.
- [ ] The payment credential equals the configured 28-byte PLB script hash exactly.
- [ ] The PLB script hash is not padded, represented as a public key, or hashed again.
- [ ] The stake credential equals Blake2b-224 of the decoded supplied public-key bytes.
- [ ] The returned address uses the network selected by `network_identifier`.
- [ ] `metadata.staking_credential` is rejected rather than ignored and returns the selected documented error.
- [ ] Wrong-case `address_type: "cip113"` returns existing error `4016`.
- [ ] Missing or malformed PLB configuration follows the documented behavior.
- [ ] No registry lookup, deployment discovery, or CIP-113 policy validation occurs.
- [ ] Existing `Enterprise`, `Base`, and `Reward` behavior remains unchanged.

#### `Cip113DeriveTest`

- The fixed public-key/PLB vector returns the expected address shown above.
- Decode the address and independently verify its address type, network, payment credential, and stake credential.
- Two different public keys produce different smart-wallet addresses.
- Identical public key, PLB, and network inputs produce the same address.
- Changing the configured PLB changes the payment credential and address.
- Mainnet returns an `addr1...` address with the mainnet network tag.
- Preview/Preprod returns an `addr_test1...` address with the testnet network tag.

#### `Cip113DeriveValidationTest`

- Missing public key follows existing error behavior.
- Invalid hexadecimal public-key bytes are rejected.
- Unsupported public-key lengths are rejected.
- Unsupported curve types are rejected; `edwards25519` is accepted.
- Existing supported public-key forms remain consistent with Enterprise derivation.
- A supplied `metadata.staking_credential` is rejected with the selected documented error.
- Missing, empty, non-hexadecimal, short, and long PLB configuration is rejected according to the chosen startup/request behavior.
- Wrong-case `"cip113"` returns existing error `4016`.

#### Regression tests

- Existing `Enterprise` derivation is unchanged.
- Existing `Base` derivation is unchanged.
- Existing `Reward` derivation is unchanged.
- Omitting `metadata.address_type` continues to select the existing default.

Kartiiyer12:
## Summary

- add the `CIP-113` address type to `/construction/derive`
- derive a smart-wallet base address from the configured PLB script credential and the supplied public-key-derived stake credential
- use cardano-client-lib's typed `Credential` API so address-kind and network header selection are not hardcoded locally
- validate the PLB as a 56-character hex string with `HexUtils.isHexString` before decoding, following the `PolicyIdValidator` pattern; validate the CIP-113 public key the same way (64 or 128 hex characters) so no decode is wrapped in try/catch
- use `198ec641705835b5e9664d0c8214a676f121e2b0d5ab8a1ecbc0ed38` for the configured Preview PLB value and matching test vectors
- validate CIP-113 public-key format, curve type, staking-credential usage, and PLB configuration without changing the shared Enterprise/Base/Reward key path
- preserve existing Enterprise, Base, Reward, and default derivation behavior

## PR boundary

This PR contains production code, deployment configuration, API documentation, compatibility expectations, and unit tests.

Integration and golden-test coverage is isolated in stacked PR #786, which uses this branch as its base.

The broader shared derive public-key validation improvement is isolated in stacked PR #787.

## Release targeting

This PR targets `develop2` because CIP-113 is scheduled for the release after the release currently being prepared on `develop`.

## Validation

- the GitHub build completed the `api` module successfully with 1,023 tests, zero failures, zero errors, and eight skips
- the full Maven reactor subsequently failed in `yaci-indexer` because GitHub Packages returned `401 Unauthorized` while resolving the Yaci Store snapshot dependencies
- the integration workflow failed during environment setup, before tests ran, because `/home/runner/.m2/repository` was unavailable as the Docker build context
- development and stacked-test diffs pass `git diff --check`
- all changed integration JSON fixtures parse successfully
- explicit unit coverage includes mainnet and test-network derivation

Closes #768


api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java
```diff
@@ -597,13 +603,64 @@ public String getCardanoAddress(AddressType addressType,
...
+            .map(String::trim)
+            .orElse("");
+
+    if (scriptHash.isEmpty()) {
+      throw ExceptionFactory.cip113PlbScriptHashNotConfigured();
+    }
+    byte[] scriptHashBytes;
+    try {
+      scriptHashBytes = decodeHexString(scriptHash);
+    } catch (RuntimeException exception) {
```
matiwinnetou: We don't usually catch runtime exception in java, you can catch Exception class

matiwinnetou: and even then you could check if these are hex bytes differently, e.g. HexUtils isHex or something like this, maybe apache commons maybe some internal utility or simply regex. What you doing here is a bit a smell: like using exceptions for control flow.

Kartiiyer12: Addressed in f33054e11. Both validations now check the input explicitly first — `length == 56 && HexUtils.isHexString(s)` for the PLB, following the existing `PolicyIdValidator` pattern, and 64/128 hex chars for the public key — and only then decode. With the input proven well-formed there is nothing to catch, so the try/catch is gone from both sites rather than retyped. No behaviour change; the 5063/5064/4007 tests are unchanged and pass.

Kartiiyer12: Addressed in f33054e11. Both validations now check the input explicitly first — `length == 56 && HexUtils.isHexString(s)` for the PLB, following the existing `PolicyIdValidator` pattern, and 64/128 hex chars for the public key — and only then decode. With the input proven well-formed there is nothing to catch, so the try/catch is gone from both sites rather than retyped. No behaviour change; the 5063/5064/4007 tests are unchanged and pass.

.env.docker-compose
```diff
@@ -131,3 +131,5 @@ PEER_DISCOVERY=false
 TOKEN_REGISTRY_ENABLED=false
 # by default fetching of logo is disabled as it can add up to a lot of bytes over the wire
 TOKEN_REGISTRY_LOGO_FETCH=false
+
+CIP113_BASE_SCRIPT_HASH=
```
matiwinnetou, line 135: before we go live, these values have to be fixed

.env.docker-compose-preprod
```diff
@@ -129,3 +129,5 @@ PEER_DISCOVERY=true
 ## Token Registry
 TOKEN_REGISTRY_ENABLED=true
 TOKEN_REGISTRY_LOGO_FETCH=true
+
+CIP113_BASE_SCRIPT_HASH=
```
matiwinnetou, line 133: before we go live, these values have to be fixed

matiwinnetou:
@Kartiiyer12 - 

1/ there are failures to build it?
2/ can you add python integration tests in the same PR?

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java
```diff
@@ -597,13 +606,60 @@ public String getCardanoAddress(AddressType addressType,
...
+      throw ExceptionFactory.cip113PlbScriptHashNotConfigured();
+    }
+    if (scriptHash.length() != SCRIPT_HASH_HEX_LENGTH || !HexUtils.isHexString(scriptHash)) {
+      throw ExceptionFactory.cip113PlbScriptHashInvalid();
+    }
+
+    return decodeHexString(scriptHash);
+  }
+
+  private HdPublicKey getValidatedCip113PublicKey(PublicKey publicKey) {
```
matiwinnetou, line 645: Why are we assuming that we will always have HD key for this, what if the public key is a simple one not representing HD wallet?

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/ConstructionApiServiceImpl.java
```diff
@@ -79,6 +79,10 @@ public ConstructionDeriveResponse constructionDeriveService(
     if (addressType == AddressType.BASE) {
       stakingCredential = Optional.ofNullable(metadata.getStakingCredential())
               .orElseThrow(ExceptionFactory::missingStakingKeyError);
+    } else if (addressType == AddressType.CIP_113) {
+      if (metadata.getStakingCredential() != null) {
```
matiwinnetou: style: you could put this second condition in the if statement itself, no need to nest it?

api/src/test/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImplTest.java
```diff
@@ -389,6 +395,160 @@ void getCardanoNullAddressTest() {
...
+    String expectedAddress = cardanoService.getCardanoAddress(AddressType.CIP_113, null,
+        givenPublicKey(), PREPROD);
+    String actualAddress = cardanoService.getCardanoAddress(AddressType.CIP_113, null,
+        publicKeyWithoutCurve, PREPROD);
+
+    assertEquals(expectedAddress, actualAddress);
+  }
+
+  @Test
+  void getCardanoCip113Address_whenCurveIsUnsupported_thenThrowsInvalidPublicKeyFormat() {
```
matiwinnetou, line 540: In the unit tests, you only seem to have negative cases, what about a happy path or am I missing something?

.env.docker-compose
Kartiiyer12: Yeah, hence added these as empty as of now so that it can fixed later.

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImpl.java
Kartiiyer12: We're not assuming HD , the endpoint explicitly accepts both a plain (non-HD) 32-byte ed25519 key and an HD-style 64-byte extended key.

getValidatedCip113PublicKey validates the input is either RAW_KEY_HEX_LENGTH (64 hex chars = 32 bytes, a bare ed25519 pubkey with no HD lineage) or EXTENDED_KEY_HEX_LENGTH (128 hex chars = 64 bytes, pubkey + BIP32 chain code), then delegates to the existing shared getHdPublicKeyFromRosettaKey, the same method already used for Enterprise/Base/Reward derivation.

api/src/test/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImplTest.java
Kartiiyer12: Happy Paths  are above these tests [here](https://github.com/cardano-foundation/cardano-rosetta-java/blob/feat/cip-113-derive-768/api/src/test/java/org/cardanofoundation/rosetta/api/construction/service/CardanoConstructionServiceImplTest.java#L399-L443)

api/src/main/java/org/cardanofoundation/rosetta/api/construction/service/ConstructionApiServiceImpl.java
Kartiiyer12: Yes, changed it.

PR #788, opened 2026-09-16
ci: add prerelease verification workflows

linconvidal:
Adds automated prerelease verification across Preprod, Mainnet Compose, and Mainnet K3s.


matiwinnetou (approved):
We are merging this for now.

Issue #770, opened 2026-07-14
add 'resolve_smart_wallet_addr' in the /network/options response


Kartiiyer12:
### Background
The /network/options endpoint in the response outputs the call methods available. So the new method needs to be added. 

``` json

Response changes - 
        "call_methods": [
            "get_parse_error_blocks",
            "mark_parse_error_block_checked",
            "resolve_smart_wallet_addr"
        ],

```

Issue #769, opened 2026-07-14
2. `/call` — `resolve_smart_wallet_addr` Method - CIP-113


Kartiiyer12:
### Background

The `/call` endpoint is used for Cardano-specific RPC methods not in the Rosetta spec.
CIP-113 adds `resolve_smart_wallet_addr` — a utility method that accepts any Cardano address
(enterprise, base, or already a smart wallet) and returns the corresponding CIP-113 smart
wallet address. This is idempotent: passing a smart wallet returns itself.

Exchanges use this before construction to convert a user's standard withdrawal address into
the address where their PLTs actually live.

### How It Works (Address Extraction Logic)

Input address type | How to extract user credential | Result
-- | -- | --
Enterprise (addr_test1v...) | Payment credential (key hash) → becomes stake/user cred of smart wallet | BaseAddress(PLB, userKeyHash)
Base (addr_test1q...) | Stake credential (key hash) used directly as user cred | BaseAddress(PLB, stakeCredential)
Smart wallet (addr_test1z...) | Stake credential extracted and re-derived (idempotent) | Same smart wallet unchanged
Script enterprise (addr_test1w..., script ≠ PLB) | Credential IS extractable but Phase 1 does not support script-owned smart wallets | Error 
Stake/Reward address | No payment credential to use as user cred | Error 

### Request / Response Examples

**Enterprise address input:**

```json
// REQUEST
{
  "network_identifier": { "blockchain": "cardano", "network": "preprod" },
  "method": "resolve_smart_wallet_addr",
  "parameters": {
    "address": "addr_test1vqa8vla3xc6kje8fflgjmzakfmme5y6v7xp7q4dhtky0gycwt8fvw"
  }
}

// RESPONSE
{
  "result": {
    "account_identifier": {
      "address": "addr_test1zreps2cq5daaw3hzq46untcp4vcnzgsn29xdx8589c9z50h67l5jmdqp66ezy7t607qnt6u08548k3v86t86x850zv2q8hv2dj"
    }
  },
  "idempotent": true
}
```

**Base address input:**

```json
// REQUEST: addr_test1qz8nzrmel9mmqnyrr6r2qv6zyras5pk0h5wfgay4agmzq4cqzppp5sllda3lrk5k7txy6kg5w2afqpfh2spwxxjc7t2asnt0u8p
// RESPONSE: same smart wallet (uses STAKE credential of base address)
{
  "result": {
    "account_identifier": {
[...]

PR #786, opened 2026-09-15
test: add CIP-113 derive integration coverage

Kartiiyer12:
## Summary

- add Spring-backed integration coverage for CIP-113 `/construction/derive`
- add Devkit golden examples for successful derivation, omitted-curve compatibility, and invalid public-key inputs
- add network-independent golden examples for rejected `metadata.staking_credential` (`5062`) and wrong-case `"cip113"` (`4016`)
- independently decode the derived address in unit coverage and verify its base-address type, network ID, exact PLB payment credential, and Blake2b-224 user stake credential
- cover the supported 64-byte extended public-key form, missing public key, deterministic output, public-key changes, and PLB changes
- update the `/network/options` golden response for the newly advertised errors
- configure the integration-test profile with PLB hash `198ec641705835b5e9664d0c8214a676f121e2b0d5ab8a1ecbc0ed38`

## Stack

This PR is stacked on development PR #785 (`feat/cip-113-derive-768`). Review and merge #785 first.

It is a test-only PR containing integration tests, golden fixtures, test configuration, and focused unit coverage. It contains no production-code changes.

## Fixture gating

- configured-PLB success and invalid-public-key goldens are tagged `devkit`, where the integration environment supplies the PLB hash
- staking-credential and wrong-case validation goldens run on every network because those requests are rejected before PLB configuration is consulted
- missing and malformed PLB behavior remains covered by unit tests; no Preprod-only golden is claimed because the Preprod workflow does not run for this stacked PR or for PRs targeting `develop2`
- the original Enterprise invalid-hex fixture remains unchanged and continues to expect its established `5000` behavior

## Validation

- all changed JSON fixtures parse successfully
- the stacked diff passes `git diff --check`
- the expected Bech32 vector independently decodes to header `0x10`, the exact configured PLB payment credential, and the Blake2b-224 hash of the supplied public key as the stake credential
- the local Maven run at commit `8b6135ee8` completed with 1,026 tests, zero failures, zero errors, and eight skips; current head `2cb74d254` adds one missing-public-key unit test, so exact-head CI verification is pending
- GitHub's integration workflow is currently blocked during environment setup, before golden execution, because `/home/runner/.m2/repository` is unavailable as the Docker `host-m2` build context

Test coverage for #768; development PR #785 closes the issue.

PR #789, opened 2026-09-16
feat: resolve CIP-113 smart wallet addresses via /call

Kartiiyer12:
## Summary

- add `resolve_smart_wallet_addr` to `/call`: accepts a key Enterprise, key/key Base, or existing CIP-113 smart-wallet address and returns the smart-wallet address for the same user credential
- Enterprise supplies its payment key hash; Base and smart-wallet addresses supply their stake key hash, so an Enterprise and a Base address for the same user resolve to the same wallet
- extract the PLB configuration check and script-payment/key-stake address construction from `CardanoConstructionServiceImpl` into a shared `Cip113AddressService`, preserving the typed-credential construction and 28-byte validation from #785, so `/construction/derive` and `/call` share one implementation
- validation order: parameter → PLB configuration → era → Shelley parse → canonical prefix → network → credential type, so the configuration error takes precedence over malformed input as it does for derive
- reuse `5051` for a missing, blank, or non-string `address`; add `5065` (payment script is not the configured PLB, including script Enterprise), `5066` (reward, pointer, Byron, script-stake), `5067` (invalid encoding, non-canonical prefix, network mismatch); inherit `5063`/`5064`
- advertise the method through `getSupportedMethods()`, which feeds `/network/options` `call_methods` — this closes #770

## Review guide

**Production code — two files carry the logic:**

- `api/src/main/java/org/cardanofoundation/rosetta/api/call/service/CallServiceImpl.java` — `resolveSmartWalletAddress` and its helpers. The credential table in the Summary maps directly onto `resolveUserCredential` / `resolveEnterpriseCredential` / `resolveBaseCredential`. `requireCanonicalPrefix` rejects an HRP that disagrees with the header byte, which the network check alone would miss on a mainnet node.
- `api/src/main/java/org/cardanofoundation/rosetta/common/services/Cip113AddressServiceImpl.java` — the shared primitive. `getConfiguredScriptHash()` is #785's validation moved verbatim; `buildSmartWalletAddress` is #785's `AddressProvider.getBaseAddress(Credential.fromScript, Credential.fromKey, network)` moved verbatim.

**Input handling worth a deliberate look:** a bech32 string under a Shelley HRP whose header nibble is the Byron marker (0x8) passes the checksum and the era check but has no HRP in cardano-client-lib, so the canonical-prefix check would throw. It is rejected with 5066 immediately after parsing (); the regression vector is in .

**Input handling worth a deliberate look:** a bech32 string under a Shelley HRP whose header nibble is the Byron marker (0x8) passes the checksum and the era check but has no HRP in cardano-client-lib, so the canonical-prefix check would throw. It is rejected with 5066 immediately after parsing (`8f12fbd0b`); the regression vector is in `CallServiceImplTest`.

**Decisions worth a deliberate look:**

- Script Enterprise returns **5065 even when its script equals the PLB** — an Enterprise address has no delegation part, so there is no user key credential to extract regardless of the payment script.
- Configuration is checked before the address, matching #785's F6 decision for derive. On an unconfigured node a malformed address therefore returns 5063, not 5067. `getCardanoCip113Address_whenPlbMissingAndPublicKeyInvalid_thenThrowsNotConfigured` in #785 and `configurationIsCheckedBeforeAddress` here pin the same precedence.
- `4015 INVALID_ADDRESS` is deliberately not reused for 5067: it is declared `retriable: true`, and changing that would alter shared behaviour on other endpoints.
- `CardanoConstructionServiceImpl.getCip113Address` now delegates to the service; `getValidatedCip113PublicKey` (curve/hex/length → 4007) stays there because it is derive-specific. Existing derive tests are the regression guard for the extraction.

**Tests — `api/src/test/java/org/cardanofoundation/rosetta/api/call/service/CallServiceImplTest.java`** (new; no test class for `CallServiceImpl` existed before). 33 tests, no mocks — the resolver depends only on `Cip113AddressServiceImpl`. `Resolution` covers each input type on both networks and decodes one result per network to assert the structure directly rather than by string comparison. Fixture addresses are checksum-valid and derived from the Preview PLB, so `SMART_WALLET_USER` is the address actually deployed on Preview.

## PR boundary

This PR contains production code, API documentation, compatibility expectations, and unit tests.

Integration and golden-test coverage for `/call` is in stacked PR #790, which uses this branch as its base.

## Release targeting

Stacked on #785 (`feat/cip-113-derive-768`) and targets it directly; it will retarget to `develop2` when #785 merges.

## Validation

- `api` module: 1,057 tests, zero failures, zero errors, eight skips — #785's 1,023 plus 34 new
- 18-case live smoke matrix against a running online instance on the rebased build: all pass, every successful result independently decoded (header `0x10`, payment credential equals the configured PLB, stake credential equals the credential extracted from the input); the operator-supplied Preview smart wallet resolves to itself, and the Enterprise address for that user resolves to it
- `/construction/derive` output unchanged after the extraction; `/network/options` advertises the method and all seven CIP-113 error codes
- `git diff --check` clean

GitHub CI on this PR inherits the `develop2` base's unresolvable yaci-store snapshot: the build job passes the `api` module and then fails in `yaci-indexer`; the integration job fails at environment setup. #786 carries the fix (`dfe16c5f2`); both jobs pass there.

## Observation for a follow-up

`CallApiImplementation` rejects every `/call` method in offline mode, a guard that predates this work because the two existing methods need the database. `resolve_smart_wallet_addr` is a pure function of the address bytes and configuration — the same computation `/construction/derive` performs, and derive is served offline. An exchange running an offline signing host can derive but not resolve. Worth deciding whether the guard should become per-method.

Closes #769
Closes #770

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Lhkk7Y8B8HuzQomvHVpmUg


api/src/main/java/org/cardanofoundation/rosetta/api/call/service/CallServiceImpl.java
```diff
@@ -42,11 +55,155 @@ public CallResponse processCallRequest(CallRequest callRequest) {
...
+        NetworkEnum network = NetworkEnum.findByName(callRequest.getNetworkIdentifier().getNetwork())
+                .orElseThrow(ExceptionFactory::invalidNetworkError);
+        byte[] configuredScriptHash = cip113AddressService.getConfiguredScriptHash();
+
+        EraAddressType eraAddressType = CardanoAddressUtils.getEraAddressType(inputAddress);
+        if (eraAddressType == null) {
+            throw ExceptionFactory.cip113InvalidAddress(
+                    "The provided address is malformed or has an unknown era");
+        }
+        if (eraAddressType == EraAddressType.BYRON) {
```
matiwinnetou, line 76: What about icarus addresses, do we want to check for those as well?

api/src/main/java/org/cardanofoundation/rosetta/api/call/service/CallServiceImpl.java
```diff
@@ -42,11 +55,155 @@ public CallResponse processCallRequest(CallRequest callRequest) {
...
+        // has no HRP for that type, so it must be rejected before the canonical-prefix check.
+        if (address.getAddressType() == AddressType.Byron) {
+            throw ExceptionFactory.cip113AddressTypeNotSupported(
+                    "Byron addresses are not supported");
+        }
+
+        requireCanonicalPrefix(address);
+
+        if (address.getNetwork().getNetworkId() != network.getNetwork().getNetworkId()) {
+            throw ExceptionFactory.cip113InvalidAddress(
```
matiwinnetou, line 92: Remind me if in rosetta we have 4xx category of errors. can you check if this is 4xx, I remember Duc did some work on this where repeatable were 5xx and non repeatable were 4xx? these "exceptions" that you are throwing is 4xx case because passing the same broken address won't result in this working... (user request error)

api/src/main/java/org/cardanofoundation/rosetta/api/call/service/CallServiceImpl.java
```diff
@@ -42,11 +55,155 @@ public CallResponse processCallRequest(CallRequest callRequest) {
...
+                    "Address network does not match requested network '%s'".formatted(network.getName()));
+        }
+
+        byte[] userCredential = resolveUserCredential(address, configuredScriptHash);
+        String smartWalletAddress = cip113AddressService.buildSmartWalletAddress(
+                configuredScriptHash, userCredential, network);
+
+        Map<String, Object> accountIdentifier = new LinkedHashMap<>();
```
matiwinnetou, line 100: I don't remembee this ouf top of my head but can you check if we don't have some generated strongly typed object for such things, we could look into other areas where are doing populating such an address object in rosetta. Worth a check.

api/src/main/java/org/cardanofoundation/rosetta/common/services/Cip113AddressServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+  private static final int SCRIPT_HASH_HEX_LENGTH = 56;
+
+  @Value("${cardano.rosetta.CIP113_BASE_SCRIPT_HASH:}")
+  private String cip113BaseScriptHash;
+
+  @Override
+  public @NotNull byte[] getConfiguredScriptHash() {
+    String scriptHash = Optional.ofNullable(cip113BaseScriptHash)
+        .map(String::trim)
+        .orElse("");
```
matiwinnetou, line 30: cosmetic: you could just do: orElseThrow(() -> {throw ExceptionFactory.cip113PlbScriptHashNotConfigured(); }) I guess....

